PR #27, opened 2020-11-13
Support for relocated docker root directory/container log files

gsmith-sas:


logging/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,44 @@
...
+2. Replace the `mountPath` and `hostPath: path` values with the location to which your Docker root directory has been mapped.
+```bash
+- mountPath: /<my_Docker_root_directory>/containers
+
+...
+
+- hostPath:
+    path: /<my_Docker_root_directory>/containers
+```
+3. Redeploy logging using the standard deployment script:
```
gsmith-sas, line 40: They do not need to redeploy all of logging, only the Fluent Bit pods.  So this bullet point should read:

3. Redeploy the Fluent Bit pods to pick-up the change by running the Fluent Bit deployment script:
```bash
/path/to/this/repo/logging/bin/deploy_logging_fluentbit_open.sh
```

logging/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,44 @@
+# Troubleshooting
+
+## Issue: Fluent Bit runs without errors, but no logs are produced
+
+### Description
+If you move the Docker root directory in your Kubernetes cluster, symbolic links are created to the 
+directory from which Fluent Bit reads log files. However, Fluent Bit cannot read the files in the directory, 
+which are present only as links. Because Fluent Bit can see the directory, it does not issue an error 
+message, but it does not produce any logs to be displayed in Kibana.
```
gsmith-sas, line 9: replace "produce" with "collect" or, maybe with "process".  They key point is that Fluent Bit is not _producing_ the log files in this location, it is reading the log messages and sending them on to Elasticsearch (to be surfaced in Kibana).

gsmith-sas (commented):
@davidbass13 We can talk through any of this if it isn't clear.  I think we may need to think about adding a fuller explanation of the "USER_DIR" concept somewhere.  I know we need to talk about clarifying the use of the "samples" as well; so maybe that's all one conversation.  In any case, that's outside the scope of the Pull Request.

logging/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Solution
+1. Edit `$USER_DIR/samples/generic-base/logging/user-values-fluent-bit-open.yaml` and uncomment these lines:
```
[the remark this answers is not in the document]
davidbass13: I did consider how much of the USER_DIR stuff I needed to include. My thinking in leaving it out was that if they were encountering this problem, they would have already deployed logging and found that it didn't work, so they would have also already set the USER_DIR. But maybe that's a bad assumption - maybe they'll look at Troubleshooting before deployment. So I'll add the step as suggested.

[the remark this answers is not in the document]
davidbass13: I used and revised the text from the esmulti sample, which reverses the two steps:

1. Copy the `../samples/generic-base` sample directory (including sub-directories) to a separate local path.

See the [main README](../../README.md#customization) for information about the customization process.

2. Set the `USER_DIR` environment variable to the local path:

```bash
export USER_DIR=/path/to/my/copy/generic-base
```

logging/TROUBLESHOOTING.md
davidbass13: Changed to "process". I didn't think "produce" was correct, but I wasn't sure what the correct action was.

BryanEllington (approved):
Simple enough

PR #31, opened 2020-11-25
Logdeploysplit

gsmith-sas:
Refactor: Monolithic Logging Deployment/Removal script into per-component scripts


logging/bin/deploy_fluentbit_azmonitor.sh
```diff
@@ -61,13 +61,21 @@ fi
...
+
 # Create ConfigMap containing Fluent Bit configuration
 kubectl -n $LOG_NS apply -f $FB_CONFIGMAP
 
 # Create ConfigMap containing Viya-customized parsers (delete it first)
 kubectl -n $LOG_NS delete configmap fbaz-viya-parsers --ignore-not-found
 kubectl -n $LOG_NS create configmap fbaz-viya-parsers  --from-file=logging/fb/viya-parsers.conf
 
+# TO DO: Replace with Helm release removal?
 # Delete any existing Fluent Bit pods in the $LOG_NS namepace (otherwise Helm chart may assume an upgrade w/o reloading updated config
```
BryanEllington, line 79: Since the pods are managed as a daemonset, deleting the pods won't do much, right?

logging/bin/deploy_kibana_content.sh
```diff
@@ -0,0 +1,160 @@
...
+  # Get first node
+  NODE_NAME=$(kubectl get nodes | awk 'NR==2 { print $1 }')
+fi
+
+
+# Need to wait 2-3 minutes for kibana to come up and
+# and be ready to accept the curl commands below
+# wait for pod to show as "running" and "ready"
+
+log_info "Checking status of Kibana pod"
```
BryanEllington, line 67: log_debug?

logging/bin/deploy_kibana_content.sh
```diff
@@ -0,0 +1,160 @@
...
+
+
+log_info "Configuring Kibana"
+
+#### TEMP?  Remove when defining nodePort via HELM chart?
+SVC=v4m-es-kibana-svc
+SVC_TYPE=$(kubectl get svc -n $LOG_NS $SVC -o jsonpath='{.spec.type}')
+if [ "$SVC_TYPE" == "NodePort" ]; then
+  KIBANA_PORT=31033
+  kubectl -n "$LOG_NS" patch svc "$SVC" --type='json' -p '[{"op":"replace","path":"/spec/ports/0/nodePort","value":31033}]'
```
BryanEllington, line 49: Why aren't we using the helm chart for this?

logging/bin/deploy_esexporter.sh
```diff
@@ -0,0 +1,94 @@
...
+# check for pre-reqs
+
+# Confirm namespace exists
+if [ "$(kubectl get ns $LOG_NS -o name 2>/dev/null)" == "" ]; then
+  log_error "Namespace [$LOG_NS] does NOT exist."
+  exit 98
+fi
+
+# get credentials
+if [ "$ES_METRICGETTER_USER" == "" ] || [ "$ES_METRICGETTER_PASSWD" == "" ] ; then
```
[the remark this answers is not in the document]
gsmith-sas: I wanted the identities to have names that reflected their purpose rather than any particular technology currently being used.  I didn't think "exporter" made the purpose/role of the identity clear.  It makes sense if you know that the Elasticsearch Exporter is the thing that collects metrics, but not otherwise.  If we later move to some other tool for metric collection, the "metric getter" identity would still make (some) sense.  I used similar logic (?) when deciding the the identity used by Fluent Bit was the "Log Collector" identity rather than "Fluent Bit User".  If we later support Logstash or FileBeats as the log collector component, we can continue to use the "Log Collector" identity there as well.

logging/bin/deploy_esexporter.sh
```diff
@@ -0,0 +1,94 @@
...
+source logging/bin/common.sh
+
+this_script=`basename "$0"`
+
+log_debug "Script [$this_script] has started [$(date)]"
+
+ELASTICSEARCH_EXPORTER_ENABLED=${ELASTICSEARCH_EXPORTER_ENABLED:-true}
+
+if [ "$ELASTICSEARCH_EXPORTER_ENABLED" != "true" ]; then
+  log_warn "Environment variable [ELASTICSEARCH_EXPORTER_ENABLED] is not set to 'true'; exiting WITHOUT deploying Elasticsearch Exporter"
```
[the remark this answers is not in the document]
gsmith-sas: Even if you've set an environment variable to disable a feature and then call a script that's primary purpose is deploy that same feature?  The wrapper/driver script treats the flag being set to "false" as a informational thing, emits an INFO message and does NOT call this script.  But, if this script was called explicitly, it seemed like we might want to highlight the disconnect.  But I'm not opposed to changing that.  We should probably have a discussion about coming up with a standard approach to handling these sorts of patterns (e.g. wrapper scripts, flags, etc.).

logging/bin/deploy_esexporter.sh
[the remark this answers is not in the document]
BryanEllington: How about ES_METRICS_USER, ES_METRICS_PASSWD?

logging/bin/deploy_esexporter.sh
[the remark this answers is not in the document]
BryanEllington: Yes. Invoking the script may be part of some backend automation and everything may be fine. Maybe use log_info instead of log_debug.

logging/bin/deploy_fluentbit_azmonitor.sh
gsmith-sas: The delete pods logic was added because the configuration is stored in a configMap and changes made to that configMap didn't trigger pods to be re-created when the Helm chart was redeployed (from Helm's perspective, nothing changed...so, no need to recreate the pods).  Deleting the pods forced them to re-read the configMap.  The "TO DO" note is suggesting a different approach of always undeploying the Helm release and re-deploying it.  That may be more attractive since the current behavior results in an extra round of pod deletions/recreations if there are changes that Helm does recognize.  The current approach works but looks "odd" if you're monitoring the pods and probably results in some wasted time/effort.

BryanEllington: Oh, then deleting the pods is better. Just remove the comment about removing the helm release. That sounds like a terrible way to go.

logging/bin/deploy_elasticsearch_open.sh
```diff
@@ -0,0 +1,291 @@
...
+   fi
+done
+
+if [ "$podready" != "TRUE" ]; then
+   log_error "The Elasticsearch pod [v4m-es-master-0] has NOT reached [Ready] status in the expected time; exiting."
+   log_error "Review pod's events and log to identify the issue and resolve it; run the remove_logging.sh script and try again."
+   exit 1
+fi
+
+# TO DO: Convert to curl command to detect ES is up?
```
BryanEllington, line 263: This is fine for now. Maybe we move away from fixed sleeps in the future to speed things up in the future as long as we have valid gate checks.

logging/bin/deploy_kibana_content.sh
```diff
@@ -0,0 +1,187 @@
...
+get_credentials_from_secret admin
+rc=$?
+if [ "$rc" != "0" ] ;then log_info "RC=$rc"; exit $rc;fi
+
+set -e
+
+
+log_info "Configuring Kibana"
+
+#### TEMP?  Remove when defining nodePort via HELM chart?
```
BryanEllington, line 53: This isn't a question - when the helm chart supports nodePort, we'd move to using that.

BryanEllington (approved):
Looks good

logging/bin/deploy_elasticsearch_open.sh
```diff
@@ -0,0 +1,291 @@
...
+export ES_LOGCOLLECTOR_PASSWD=${ES_LOGCOLLECTOR_PASSWD:-$(uuidgen)}
+export ES_METRICGETTER_PASSWD=${ES_METRICGETTER_PASSWD:-$(uuidgen)}
+
+# Create secrets containing internal user credentials
+create_user_secret internal-user-admin        admin        $ES_ADMIN_PASSWD         managed-by=v4m-es-script
+create_user_secret internal-user-kibanaserver kibanaserver $ES_KIBANASERVER_PASSWD  managed-by=v4m-es-script
+create_user_secret internal-user-logcollector logcollector $ES_LOGCOLLECTOR_PASSWD  managed-by=v4m-es-script
+create_user_secret internal-user-metricgetter metricgetter $ES_METRICGETTER_PASSWD  managed-by=v4m-es-script
+
+# Print info about how to obtain admin password
```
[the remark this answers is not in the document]
gsmith-sas: Latest update only displays the message when it has been generated for the user.

PR #32, opened 2020-12-01
Patch KubeHpaMaxedOut alert to check for > 1 replicas

BryanEllington:


monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -181,6 +181,9 @@ for f in monitoring/rules/viya/rules-*.yaml; do
   kubectl apply -n $MON_NS -f $f
 done
 
+log_info "Patching KubeHpaMaxedOut rule..."
+kubectl patch prometheusrule --type='json' -n monitoring v4m-kubernetes-apps --patch "$(cat monitoring/kube-hpa-alert-patch.json)"
```
gsmith-sas, line 185: Hard-coded namespace?  

Probably need/want to add an explanation or some details about the change (i.e. "_limits false alerts triggered by..._").  Can't have comments in the patch JSON file and the contents of the patch file are fairly opaque.

gsmith-sas (changes requested):
The hard-coded namespace almost certainly needs to be fixed...the expanded comment is a suggestion.

gsmith-sas (approved):

PR #34, opened 2020-12-07
Randomly generate Grafana password by default

BryanEllington:


monitoring/README.md
```diff
@@ -187,7 +193,10 @@ applications are available at these locations by default:
 * Prometheus - Port 31090 `http://master-node.yourcluster.example.com:31090`
 * AlertManager - Port 31091 `http://master-node.yourcluster.example.com:31091`
 
-The default credentials for Grafana are `admin`:`admin`.
+The default Grafana admin user is `admin`. Unless otherwise specified,
```
gsmith-sas, line 196: Is the default "admin" or randomly-generated?  The first two sentences seem to contradict themselves.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -114,16 +114,25 @@ if [ "$TLS_ENABLE" == "true" ]; then
...
 fi
+log_info "User response file: [$PROM_OPER_USER_YAML]"
+log_info "Deploying the Kube Prometheus Stack. This may take a few minutes..."
+if helm3ReleaseExists $promRelease $MON_NS; then
+  log_info "Upgrading via Helm...($(date) - timeout 20m)"
+else
+  grafanaPwd="$GRAFANA_ADMIN_PASSWORD"
+  if [ "$grafanaPwd" == "" ]; then
+    showPass="true"
+    grafanaPwd="$(date +%s | sha256sum | base64 | head -c 32 ; echo)"
```
gsmith-sas, line 132: I thought you were creating a bash function (in common.sh?) for generating the password so we'd handle things consistently b/w logging and monitoring.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -114,16 +114,25 @@ if [ "$TLS_ENABLE" == "true" ]; then
...
 else
   promRelease=v4m-prometheus-operator
   promName=v4m
 fi
+log_info "User response file: [$PROM_OPER_USER_YAML]"
+log_info "Deploying the Kube Prometheus Stack. This may take a few minutes..."
+if helm3ReleaseExists $promRelease $MON_NS; then
+  log_info "Upgrading via Helm...($(date) - timeout 20m)"
+else
+  grafanaPwd="$GRAFANA_ADMIN_PASSWORD"
```
gsmith-sas, line 129: Not sure if it matters (we haven't really worried about consistency in this area), but the comparable environment variables on the logging side are in the form **ES_** _+identitiy +_ **_PASSWD** (e.g. ES_ADMIN_PASSWD, ES_LOGCOLLECTOR_PASSWD, etc.).

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -181,16 +191,13 @@ for f in monitoring/rules/viya/rules-*.yaml; do
...
-
 echo ""
 monitoring/bin/deploy_dashboards.sh
 
 log_notice "Successfully deployed components to the [$MON_NS] namespace"
+if [ "$showPass" == "true" ]; then
+  log_notice ""
+  log_notice "Generated Grafana admin password is: $grafanaPwd"
+  log_notice "Change the password at any time by running (replace pod name and password):"
+  log_notice "kubectl exec -n $MON_NS grafana-pod-name -- bin/grafana-cli admin reset-admin-password myNewPassword"
```
gsmith-sas, line 202: I think including the name of the actual/current Grafana pod would be more helpful.  On the logging side, when we've shown commands in this type of message, they can generally be cut-and-pasted immediately to use.  In this case, no one should want to use _myNewPassword_ as their new password, so they have to edit the command anyway but still...  

  I know any pod name shown can become obsolete fairly quickly.  But, if someone has just run the deployment script, it seems unhelpful not to provide a command they can use right away.

samples/generic-base/monitoring/user.env
```diff
@@ -38,7 +38,10 @@
 # PROM_OPERATOR_CRD_VERSION=v0.43.2
 
 # Version of the kube-prometheus-stack helm chart to use
-# KUBE_PROM_STACK_CHART_VERSION=11.1.3
+# KUBE_PROM_STACK_CHART_VERSION=11.0.0
```
gsmith-sas, line 41: Reverting to an early chart version?

monitoring/grafana-datasource-es.yaml
```diff
@@ -15,5 +15,6 @@ datasources:
     logMessageField: message
     logLevelField: level
   secureJsonData:
-    basicAuthPassword: admin
+    # Change this value
+    basicAuthPassword: YOUR_PASSWORD_HERE
```
gsmith-sas, line 19: Is this value still used?  Looks like they either have 2 different ways of setting it or need to set it twice.  
If the env var is the right way (or over-rides the value here) I would comment out this line and change the comment to say "_This value isn't used, set the admin password by setting the GRAFANA_ADMIN_PASSWORD environment variable._"

monitoring/README.md
```diff
@@ -112,10 +112,16 @@ Operator and aggregates other helm charts such as Grafana and the Prometheus
...
+**Note 1:** If you are using a cloud provider, you must use ingress, rather than
 NodePorts. Use the samples in the [samples/ingress](/samples/ingress)
 area of this repository to set up either host-based or path-based ingress.
 
+**Note 2:** Although the Grafana helm chart supports a value to set the initial
+admin password, you must use the `GRAFANA_ADMIN_PASSWORD` value in
+`$USER_DIR/monitoring/user.env` to set this value. If not specified, the
+initial Grafana admin password will be randomly generated and logged during the
+initial deployment of the monitoring components.
+
```
davidbass13, line 124: What do you mean by "not specified" in user.env? Does that mean the user doesn't include the GRAFANA_ADMIN_PASSWORD line in their user.env at all? If they keep the line that's in the repo user.env but don't change it, would their password be "yourPasswordHere"?

I'm also not sure what you mean by the password is "logged" during deployment. Logged where? How will they find it?

monitoring/user.env
```diff
@@ -49,6 +40,9 @@
 # Version of the kube-prometheus-stack helm chart to use
 # KUBE_PROM_STACK_CHART_VERSION=11.0.0
 
+# Initial password of the Grafana admin user
+# GRAFANA_ADMIN_PASSWORD=yourPasswordHere
+
```
davidbass13, line 45: Is having a password in plaintext in their user.env file a security concern? If the user.env file contains GRAFANA_ADMIN_PASSWORD= with no value, would it randomly generate one?

monitoring/README.md
BryanEllington: The default 'user' is still 'admin'. The password is randomly generated by default unless specified via GRAFANA_ADMIN_PASSWORD.

monitoring/bin/deploy_monitoring_cluster.sh
BryanEllington: Yep...coming in next commit

monitoring/bin/deploy_monitoring_cluster.sh
BryanEllington: Why do you hate vowels?

monitoring/bin/deploy_monitoring_cluster.sh
BryanEllington: Yeah, you're right. I'll add some logic to find the pod name that will at least be valid as of the time the deploy script was run.

samples/generic-base/monitoring/user.env
BryanEllington: Copy-paste error. Will fix.

monitoring/grafana-datasource-es.yaml
BryanEllington: It's not used by default. Setting ELASTICSEARCH_DATASOURCE=true will deploy the Elasticsearch datasource. If enabled, they will need to set the password. The password is the Elasticsearch password, not the Grafana password.

monitoring/README.md
BryanEllington: Not specified in the environment variable sense - whether that's through $USER_DIR/user.env, $USER_DIR/monitoring/user.env, or just the terminal environment.

monitoring/user.env
BryanEllington: Env vars can be set in multiple ways. I thought the user.env was our only real doc on what could be set, so I put it there. In practice, it's up to the customer whether to put the plain text password in a file or source it into their environment manually prior to deployment.

gsmith-sas: I wonder whether we should include a comment here (and in the corresponding logging file) indicating that the environment variables tied to password can also be set via other mechanism (e.g. issuing set/export command prior to running the scripts) if users need/want to avoid having passwords set in files.  And/or, include it in the doc.  Should probably also point out that setting it elsewhere means they need to remember it or take other steps to ensure it can be retrieved later.

BryanEllington: It's a general principle. Content of user.env files ends up as environment variables. Those are settable externally, like any environment variable. They could even set something like their KUBECONFIG in the user.env to tie a given USER_DIR to a particular cluster.

I would include this as part of the overall rework of user config documentation being handled separately.

gsmith-sas (approved):
LGTM

PR #35, opened 2020-12-08
Revised Grafana password information in monitoring/README

davidbass13:


BryanEllington (commented):
I think it's important to point out that the environment variable can be set via user.env files (either $USER_DIR/user.env or (preferably for this value) $USER_DIR/monitoring/user.env.

monitoring/README.md
```diff
@@ -86,14 +86,20 @@ are present in `USER_DIR`:
...
+- Helm chart parameters: These parameters are organized into a hierarchical structure and contain mixed case characters (for example, `persistentVolume:storageClass`). These parameters are specified in `.yaml` configuration files.
+
 ### user.env
 
-The `monitoring/user.env` file contains environment variable flags that customize
-the components that are deployed or to alter some script behavior (such as to
+The `monitoring/user.env` file contains environment variables that customize
+the components that are deployed or alter some script behavior (such as to
 enable debug output). All values in `user.env` files are exported as environment
 variables available to the scripts. A `#` as the first character in a line
```
gsmith-sas, line 98: The sentence related to comments should be sometnhing like: _"Any line with a '#' as the first character will be treated as a comment and ignored."_

monitoring/README.md
```diff
@@ -193,10 +199,14 @@ applications are available at these locations by default:
 * Prometheus - Port 31090 `http://master-node.yourcluster.example.com:31090`
 * AlertManager - Port 31091 `http://master-node.yourcluster.example.com:31091`
 
-The default Grafana admin user is `admin`. Unless otherwise specified,
-the default password is randomly generated. The value is logged only
-during the initial deployment of the monitoring components and not on
-an upgrade.
+The default Grafana admin user is `admin`. Unless you used the `GRAFANA_ADMIN_PASSWORD` environment variable (either in the `user.env` file or on the command line) to specify a default password during deployment, the default password is randomly generated and displayed during deployment.
```
gsmith-sas, line 202: I think the word "set" is better than "used" (i.e. _Unless you set the 'GRAFANA_ADMIN_PASSWORD' environrment variable..._)

monitoring/README.md
```diff
@@ -86,14 +86,20 @@ are present in `USER_DIR`:
 * `monitoring/user-values-prom-operator.yaml`
 * `monitoring/user-values-pushgateway.yaml`
 
+You can modify two types of values to customize your deployment:
+- Environment variables: These variables contain uppercase characters and underscores (for example, `LOG_COLOR_ENABLE=`), and can be specified in the `user.env` file or on the command line. Specifying the variables in `user.env` is recommended, in order to maintain a consistent set of values for future deployments.
```
gsmith-sas, line 90: Environment variables do NOT have to use uppercase and underscores...they often do but that is not a requirement.  I do NOT know if every env variable we support follows that convention.  I'm not sure Bryan meant for that to get into the doc...thought it was just internal guidance to help differentiate between env vars and Helm chart parameters.  If it goes in the doc, we need to weaken the statement (perhaps: _These variables often (or generally?) contain uppercase and underscores_)

monitoring/README.md
```diff
@@ -86,14 +86,20 @@ are present in `USER_DIR`:
...
+- Helm chart parameters: These parameters are organized into a hierarchical structure and contain mixed case characters (for example, `persistentVolume:storageClass`). These parameters are specified in `.yaml` configuration files.
```
gsmith-sas, line 91: See comment above about Environment Variables.

monitoring/README.md
davidbass13: Removed information about case and characters in variables

monitoring/README.md
davidbass13: Removed information about case and characters in variables

monitoring/README.md
davidbass13: Changed to "Any line whose first character is `#` is treated as a comment and ignored."

monitoring/README.md
davidbass13: Changed to "set"

monitoring/README.md
```diff
@@ -193,10 +199,14 @@ applications are available at these locations by default:
...
-The default Grafana admin user is `admin`. Unless otherwise specified,
-the default password is randomly generated. The value is logged only
-during the initial deployment of the monitoring components and not on
-an upgrade.
+The default Grafana admin user is `admin`. Unless you used the `GRAFANA_ADMIN_PASSWORD` environment variable (either in the `user.env` file or on the command line) to specify a default password during deployment, the default password is randomly generated and displayed during deployment. 
+
+If you want to change the password, issue this command:
+```bash
+kubectl exec -n <monitoring_namespace> <grafana_pod> -c grafana -- bin/grafana-cli admin reset-admin-password myNewPassword
+```
```
davidbass13, line 206: @BryanEllington, is this the correct command to change the Grafana password?

BryanEllington: Yes, that is correct

PR #37, opened 2020-12-09
TLS script and doc improvements

BryanEllington:


bin/tls-include.sh
```diff
@@ -18,11 +18,31 @@ function verify_cert_manager {
   cert_manager_ok="true"
   if [ "$TLS_ENABLE" == "true" ]; then
```
gsmith-sas, line 19: TLS is always required for logging regardless of the value of TLS_ENABLE is true or not.  Therefore, this function probably shouldn't skip processing if TLS_ENABLE is set to "false".  Or, alternatively, perhaps s we should set TLS_ENABLE = true in the deploy_logging_open.sh script?  Thoughts?

logging/bin/deploy_logging_open.sh
```diff
@@ -21,7 +21,7 @@ LOG_KB_TLS_ENABLE=${LOG_KB_TLS_ENABLE:-false}
 LOG_NODE_PLACEMENT_ENABLE=${LOG_NODE_PLACEMENT_ENABLE:-${NODE_PLACEMENT_ENABLE:-false}}
 
 source bin/tls-include.sh
-verify_cert_manager
+verify_cert_manager $LOG_NS es-transport es-rest es-admin kibana
```
gsmith-sas, line 25: Hmmm....as part of the work being done in the logdeploysplit branch this file (deploy_logging_open.sh) will change quite a bit and this call no longer exists.  Looking at the branch now, I see that we're not calling this function at all right now.  Since those changes aren't pushed yet, I guess this change is "right"...we'll just need to be careful about the order in which the PRs are merged and add a comparable call into the right place on the other branch if this PR is merged first.

samples/tls/monitoring/README.md
```diff
@@ -8,25 +8,48 @@ All components have TLS enabled on ingress. Due to limitations in the
...
-process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq) in SAS Viya Administration documentation and specify the `monitoring` namespace.
+Generating these certificates is outside the scope of this example. However,
+you can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq)
+in SAS Viya Administration documentation and specify the `monitoring` namespace.
+
+## In-Cluster TLS
+
+The monitoring components use the following secrets for the TLS certificates
+service-to-service (east-west) traffic:
+
```
davidbass13, line 29: I would revise to "The monitoring components use these secrets for the TLS certificates that handle interactions between services:"

samples/tls/monitoring/README.md
```diff
@@ -8,25 +8,48 @@ All components have TLS enabled on ingress. Due to limitations in the
...
+* `grafana-tls-secret`
+* `prometheus-tls-secret`
+* `alertmanager-tls-secret`
+
+These secrets can be manually created, but if any
+do not exist, then [cert-manager](https://cert-manager.io/) is required to
+create the missing certificates.
+
```
davidbass13, line 37: You can manually create these certificates. If any do not exist, you must use [cert-manager](https://cert-manager.io/) to create the missing certificates.

samples/tls/monitoring/README.md
```diff
@@ -8,25 +8,48 @@ All components have TLS enabled on ingress. Due to limitations in the
...
+cert-manager is not required if TLS is disabled or if all of the TLS secrets
+exist prior to deployment.
 
 For in-cluster (east-west traffic) TLS for monitoring components,  
-[cert-manager](https://cert-manager.io/) populates these secrets that contain pod certificates. Because existing secrets are not overwritten, you can manually populate them.
+[cert-manager](https://cert-manager.io/) populates these secrets that contain
+pod certificates.
```
davidbass13, line 44: [cert-manager](https://cert-manager.io/) populates these secrets, which contain pod certificates:

samples/tls/monitoring/README.md
BryanEllington: "you must use cert-manager" is a bit misleading. The scripts will automatically use cert-manager. The user isn't required to "use" cert-manager.

davidbass13: How about "However, if any certificates do not exist, [cert-manager](https://cert-manager.io/) automatically creates the missing certificates."

gsmith-sas (commented):
I think we need some tweaks...not sure if those need to be done here or on the logging side (see comments).

samples/tls/monitoring/README.md
BryanEllington: done

samples/tls/monitoring/README.md
BryanEllington: The point we're trying to get across is that if you don't create the secrets/certificates yourself, cert-manager is required to be installed in the cluster. If the user manually creates the secrets, cert-manager is **not** required.

samples/tls/monitoring/README.md
BryanEllington: done

bin/tls-include.sh
BryanEllington: My take: Override and always set to true in the logging script.

logging/bin/deploy_logging_open.sh
BryanEllington: right...if you're going to remove the check and do it your own way, let this change go through, then your PR will just modify the file again to remove it. Shouldn't be a problem.

samples/tls/monitoring/README.md
davidbass13: OK, maybe this: 

[cert-manager](https://cert-manager.io/) is not required if you manually create these certificates. However, if you do not manually create the certificates, cert-manager must be installed in the cluster in order to create any missing certificates.

gsmith-sas: How about: _"If any of the required certificates do not exist, the deployment process will attempt to use cert-manager to generate the missing certificates.  If the required certificates do not exist and cert-manager is not available, the deployment process will fail."_

davidbass13: That sounds fine - I'm assuming that replaces the second sentence and that the sentence about manually creating the certificates remains.

bin/tls-include.sh
BryanEllington: Done

samples/tls/monitoring/README.md
gsmith-sas: Yes, that makes sense to me.  But suggest you wait for @BryanEllington before considering discussion finished.

gsmith-sas (approved):
Assuming testing goes well, I'm fine with the code changes.

samples/tls/monitoring/README.md
BryanEllington: Incorporated Greg's suggestion in the latest commit

gsmith-sas (commented):
LGTM

PR #40, opened 2020-12-14
New Kibana content: Top 10 Errors and Pod/Container-focused Dashboards

gsmith-sas:


BryanEllington (approved):
Code change is trivial. Haven't personally loaded the dashboard itself, but it passed testing, so we're good.

PR #46, opened 2020-12-15
Move to prometheus-community Helm chart for es-exporter

gsmith-sas:


logging/user-values-es-exporter.yaml
```diff
@@ -1,9 +1,9 @@
...
+# Helm Chart:     https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-elasticsearch-exporter
+# Default values: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-elasticsearch-exporter/values.yaml
 
 # image:
 #   repository: justwatch/elasticsearch_exporter
 #   tag: 1.1.0
 
 # es:
-#   uri: https://admin:admin@v4m-es-client-service.logging:9200
+#   uri: https://metricgetter:metricgetterpw@v4m-es-client-service.logging:9200
```
BryanEllington, line 9: Any way to use env vars and/or secrets for the URI?

gsmith-sas: Based on the existing comment, I was showing 'sample' values...just pushed code to update comment to reflect the default behavior of setting this using env vars.

BryanEllington (approved):

PR #54, opened 2021-01-08
Improve parsing of 1st JSON fragment of excessively long messages split due to length

gsmith-sas:


logging/fb/viya-parsers.conf
```diff
@@ -197,3 +197,17 @@
        regex         \[(?<caller>\w+)\]\s+(?<timestamp>\d{4}(\/\d{2}){2} \d{2}:\d{2}:\d{2}\.\d{3})\d{3} (?<level>[A-Z]*): (?<message>.*)
        Time_Key      timestamp
        Time_format   %Y/%m/%d %H:%M:%S.%L
+[PARSER]
+       Name          truncatedjson
+       # the first fragment of a very long JSON structured log message broken up b/c it is longer than 16K (docker limit)
+       Format        regex
+       regex         {\s*"level":\s*"(?<level>\w+)",\s*"version":\s*\d,\s*"source":\s*"(?<source>\S*?)",(?<message>(\s*"messageKey":\s*"(?<messageKey>.*?)",)?.*)
```
BryanEllington, line 204: Is the order of keys in JSON very reliable - especially across programming languages? I guess we're just solving for a specific known case for now, but ugh...

gsmith-sas: lol...yep, in this particular case, we're trying to deal with a very specific set of JSON _fragments_ for which we know the ordering of keys.

BryanEllington (approved):

PR #59, opened 2021-01-25
Revised TLS sample READMEs

davidbass13:
Second attempt at revised READMEs for TLS sample. This branch should contain only the changed files - /sample/tls/logging/README.md and sample/tls/monitoring/README.md.

### DIFF
--- a/logging/README.md
+++ b/logging/README.md
@@ -270,4 +270,21 @@ You can choose to enable users to access Elasticsearch so that they can issue qu
 ./logging/bin/es_nodeport_enable.sh
 ``` 
 
+When the script completes, a message appears that contains the URL to access Elasticsearch.
+```
+=============================================
+== Access Elasticsearch using this URL: https://<myK8snode>:<portnumber>/ ==
+=============================================
+```
+
+The message provides values for `<myK8snode>` and `<portnumber>` that are specific to your deployment.
+
 User connections to Elasticsearch always use TLS.
+
+To disable nodeport connections to Elasticsearch, run this command:
+```bash
+./logging/bin/es_nodeport_disable.sh
+```  
+
+You can use these scripts to temporarily enable user access to Elasticsearch.
+
--- a/logging/README.md
+++ b/logging/README.md
@@ -270,18 +270,20 @@ You can choose to enable users to access Elasticsearch so that they can issue qu
 ./logging/bin/es_nodeport_enable.sh
 ``` 
 
-When the script completes, a message appears that contains the URL to access Elasticsearch.
-```
+When the script completes, a message such as this appears in the console window:
+
+```text
 =============================================
-== Access Elasticsearch using this URL: https://<myK8snode>:<portnumber>/ ==
+== Access Elasticsearch using this URL: https://myK8snode:<portnumber>/ ==
 =============================================
 ```
 
-The message provides values for `<myK8snode>` and `<portnumber>` that are specific to your deployment.
+The message provides the URL address for the Elasticsearch application. The `<portnumber>` value is specific to your deployment.
 
 User connections to Elasticsearch always use TLS.
 
 To disable nodeport connections to Elasticsearch, run this command:
+
 ```bash
 ./logging/bin/es_nodeport_disable.sh
 ```  
--- a/logging/README.md
+++ b/logging/README.md
@@ -261,3 +261,13 @@ deployment process.
 * Select the __Discover__ icon in the toolbar to display the Discover page. Use
 this page to review the collected log messages. You can use the query box or
 __Add filter__ to filter the messages that are displayed.
+
+## Enable Users to Access Elasticsearch (optional)
+
+You can choose to enable users to access Elasticsearch so that they can issue queries using API calls or scripts. Run this command to enable nodeport connections to Elasticsearch:
+
+```bash
+./logging/bin/es_nodeport_enable.sh
+``` 
+
+User connections to Elasticsearch always use TLS.
--- a/samples/tls/logging/README.md
+++ b/samples/tls/logging/README.md
@@ -2,9 +2,11 @@
 
 ## Overview
 
-This sample demonstrates how to deploy logging components with TLS enabled for connections to Kibana.
+This sample demonstrates how to deploy logging components with TLS enabled.
 
-All logging components use TLS for connections between components within the cluster.
+All communication within the cluster between the logging components takes place through TLS-enabled connections.
+
+The `TLS_ENABLE` environment variable controls whether connections from the user to Kibana use TLS.
 
 ## Using This Sample
 
@@ -26,7 +28,7 @@ my_repository_path/logging/bin/deploy_logging_open.sh
 
 ### Ingress
 
-* If you are using ingress, set the environment variable `TLS_ENABLED=true` in the `user.env` file in order to use TLS for connections between the user and Kibana.  
+* If you are using ingress, specify `TLS_ENABLED=true` in the `user.env` file to use TLS for communications between the ingress object and Kibana. Connections between the ingress object and Elasticsearch always use TLS, regardless of the value of `TLS_ENABLED`.
 
 * Manually populate these secrets in the `logging` namespace (or `LOG_NS` value) **before**
 you run the deployment script:
@@ -34,16 +36,26 @@ you run the deployment script:
   * kubernetes.io/tls secret - `kibana-ingress-tls-secret`
   * kubernetes.io/tls secret - `elasticsearch-ingress-tls-secret`
 
-Generating these certificates is outside the scope of this example. However, you
-can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq) in SAS Viya Administration documentation and specify the `logging` namespace.
-
-If any of the required certificates do not exist, the deployment process attempts to use [cert-manager](https://cert-manager.io/) to generate the missing certificates.
+Use cert-manager, if available, to create these certificates. If cert-manager is not available, you must create the certificates manually by using the `kubectl -n logging create secret` command.
 
 * If you are using an ingress controller other than NGINX, modify the annotation 
-`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` in the `user-values-elasticsearch-open.yaml` file. Refer to the documentation for your ingress controller. 
+`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` as needed in the `user-values-elasticsearch-open.yaml` file. Refer to the documentation for your ingress controller. 
 
 ### Nodeports
 
-If you are using nodeports,  set the environment variable `TLS_ENABLED=true` in the `user.env` file in order to use TLS for connections between the user and Kibana. 
+* If you are using nodeports, set the environment variable `TLS_ENABLED=true` in the `user.env` file to use TLS for connections between the user and Kibana.
 
+* If users need to directly connect to Elasticsearch, you must run the `es_enable_nodeport.sh` script to enable the connection. Direct connections to Elasticsearch always use TLS, regardless of the value of the `TLS_ENABLED` variable.
+
+### Secrets for In-Cluster TLS
+
+The standard deployment script for the logging components use these secrets for the TLS certificates that
+handle interactions between components:
+
+* kubernetes.io/tls secret - `es-rest-tls-secret`
+* kubernetes.io/tls secret - `es-transport-tls-secret`
+* kubernetes.io/tls secret - `kibana-tls-secret`
+
+By default, the deployment process uses [cert-manager](https://cert-manager.io/) to generate the certificates. If cert-manager is not available, you can manually generate the certificates using the process in [Notes on Using TLS](../../logging/NOTES_ON_USING_TLS.md). If the required certificates do not exist and cert-manager is not available, the deployment process fails. cert-manager is not required
+if TLS is disabled or if all of the TLS secrets exist prior to deployment.
 
--- a/samples/tls/logging/README.md
+++ b/samples/tls/logging/README.md
@@ -28,7 +28,7 @@ my_repository_path/logging/bin/deploy_logging_open.sh
 
 ### TLS 
 
-Specify `TLS_ENABLE=true` in the `user.env` file to use TLS for communications between the ingress object and Kibana. Connections between the ingress object and Elasticsearch always use TLS, regardless of the value of `TLS_ENABLE`.
+Specify `TLS_ENABLE=true` in the `user.env` file to require TLS for connections to Kibana. Connections to Elasticsearch (when enabled) always use TLS, regardless of the value of `TLS_ENABLE`.
 
 ### Ingress
 
--- a/samples/tls/logging/README.md
+++ b/samples/tls/logging/README.md
@@ -2,22 +2,9 @@
 
 ## Overview
 
-This sample demonstrates how to deploy logging components with TLS enabled.
-For now, only ingress is covered. In-cluster TLS will be supported in the
-future.
+This sample demonstrates how to deploy logging components with TLS enabled for connections to Kibana.
 
-All components have TLS enabled on ingress. Due to limitations in the
-underlying Helm charts, some components might not have TLS enabled in-cluster.
-
-If you use this sample for HTTPS for ingress, the following secrets must be
-manually populated in the `logging` namespace (or `LOG_NS` value) **BEFORE**
-you run any of the scripts in this repository:
-
-* kubernetes.io/tls secret - `kibana-ingress-tls-secret`
-* kubernetes.io/tls secret - `elasticsearch-ingress-tls-secret`
-
-Generating these certificates is outside the scope of this example. However, you
-can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq) in SAS Viya Administration documentation and specify the `logging` namespace.
+All logging components use TLS for connections between components within the cluster.
 
 ## Using This Sample
 
@@ -37,9 +24,26 @@ my_repository_path/logging/bin/deploy_logging_open.sh
 ```
 ## Notes On Customization Values
 
-Set `MON_TLS_ENABLE=true` in the `$USER_DIR/logging/user.env` file. This variable modifies the deployment of Open Distro for Elasticsearch to be TLS-enabled.
+### Ingress
 
-Edit `$USER_DIR/logging/user-values-elasticsearch-open.yaml` and replace
-any sample hostnames with hostnames for your deployment. Specifically, you must replace
-`host.cluster.example.com` with the name of the ingress node. Often, the ingress node is the cluster master node, but your environment might be different.
+* If you are using ingress, set the environment variable `TLS_ENABLED=true` in the `user.env` file in order to use TLS for connections between the user and Kibana.  
+
+* Manually populate these secrets in the `logging` namespace (or `LOG_NS` value) **before**
+you run the deployment script:
+
+  * kubernetes.io/tls secret - `kibana-ingress-tls-secret`
+  * kubernetes.io/tls secret - `elasticsearch-ingress-tls-secret`
+
+Generating these certificates is outside the scope of this example. However, you
+can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq) in SAS Viya Administration documentation and specify the `logging` namespace.
+
+If any of the required certificates do not exist, the deployment process attempts to use [cert-manager](https://cert-manager.io/) to generate the missing certificates.
+
+* If you are using an ingress controller other than NGINX, modify the annotation 
+`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` in the `user-values-elasticsearch-open.yaml` file. Refer to the documentation for your ingress controller. 
+
+### Nodeports
+
+If you are using nodeports,  set the environment variable `TLS_ENABLED=true` in the `user.env` file in order to use TLS for connections between the user and Kibana. 
+
 
--- a/samples/tls/logging/README.md
+++ b/samples/tls/logging/README.md
@@ -26,36 +26,37 @@ my_repository_path/logging/bin/deploy_logging_open.sh
 ```
 ## Notes On Customization Values
 
+### TLS 
+
+Specify `TLS_ENABLE=true` in the `user.env` file to use TLS for communications between the ingress object and Kibana. Connections between the ingress object and Elasticsearch always use TLS, regardless of the value of `TLS_ENABLE`.
+
 ### Ingress
 
-* If you are using ingress, specify `TLS_ENABLED=true` in the `user.env` file to use TLS for communications between the ingress object and Kibana. Connections between the ingress object and Elasticsearch always use TLS, regardless of the value of `TLS_ENABLED`.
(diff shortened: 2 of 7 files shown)


samples/tls/logging/README.md
```diff
@@ -2,22 +2,9 @@
 
 ## Overview
 
-This sample demonstrates how to deploy logging components with TLS enabled.
-For now, only ingress is covered. In-cluster TLS will be supported in the
-future.
+This sample demonstrates how to deploy logging components with TLS enabled for connections to Kibana.
```
BryanEllington, line 5: It also shows TLS ingress for Elasticsearch, although I'm not sure why as we would generally like to keep Elasticsearch isolated in-cluster for security reasons. Greg?

samples/tls/monitoring/README.md
```diff
@@ -4,46 +4,10 @@
...
-**BEFORE** you deploy cluster monitoring.
-
-* kubernetes.io/tls secret - `prometheus-ingress-tls-secret`
-* kubernetes.io/tls secret - `alertmanager-ingress-tls-secret`
-* kubernetes.io/tls secret - `grafana-ingress-tls-secret`
-
-Generating these certificates is outside the scope of this example. However,
-you can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq)
-in SAS Viya Administration documentation and specify the `monitoring` namespace.
+If you enable TLS, communications between the user and the monitoring components
```
BryanEllington, line 7: Setting TLS_ENABLE only enables TLS between in-cluster monitoring components.

The confusing part is that when using NodePorts (the default), the user->monitoring component communication will also be TLS-enabled since NodePorts are a direct line to the internal components without ingress in-between.

samples/tls/monitoring/README.md
```diff
@@ -64,17 +28,44 @@ my_repository_path/monitoring/bin/deploy_monitoring_cluster.sh
...
+* kubernetes.io/tls secret - `prometheus-tls-secret`
+* kubernetes.io/tls secret - `alertmanager-tls-secret`
+* kubernetes.io/tls secret - `grafana-tls-secret`
+
+If any of the required certificates do not exist, the deployment process attempts to use [cert-manager](https://cert-manager.io/) to generate the missing
+certificates. If the required certificates do not exist and cert-manager is
+not available, the deployment process fails. cert-manager is not required
+if TLS is disabled or if all of the TLS secrets exist prior to deployment.
+
+### Limitations and Known Issues
```
BryanEllington, line 65: I believe the AlertManager is resolved. I still have work to do to implement TLS for AlertManager, but we can take out the reference to the issue, which has expired and is closed.

BryanEllington (changes requested):
I think we need to do more to disentangle TLS_ENABLE from ingress.

samples/tls/logging/README.md
gsmith-sas: Some sites may choose to make Elasticsearch visible to their users.  This would enable, for example, users to submit Elasticsearch queries via API calls or scripts using one of the query languages.  We may need/want to add a sentence or two explaining that enabling TLS for Kibana and enabling TLS for Elasticsearch are independent decision.  Hmmm, that probably ties into a higher-level discussion of whether they want to make Elasticsearch directly available or not.  We make the es_nodeport_enable.sh script available to allow nodePort access to ES since we do not enable that by default.  I wonder if that needs its own README or, probably better, adding a paragraph in the top-level Logging README file.

samples/tls/logging/README.md
```diff
@@ -2,22 +2,9 @@
...
-If you use this sample for HTTPS for ingress, the following secrets must be
-manually populated in the `logging` namespace (or `LOG_NS` value) **BEFORE**
-you run any of the scripts in this repository:
-
-* kubernetes.io/tls secret - `kibana-ingress-tls-secret`
-* kubernetes.io/tls secret - `elasticsearch-ingress-tls-secret`
-
-Generating these certificates is outside the scope of this example. However, you
-can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq) in SAS Viya Administration documentation and specify the `logging` namespace.
+All logging components use TLS for connections between components within the cluster.
```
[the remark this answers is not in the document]
gsmith-sas: Maybe reword it to say _"All communications within the cluster between the logging components is done via TLS-enabled connections."_

samples/tls/logging/README.md
```diff
@@ -37,9 +24,26 @@ my_repository_path/logging/bin/deploy_logging_open.sh
 ```
 ## Notes On Customization Values
 
-Set `MON_TLS_ENABLE=true` in the `$USER_DIR/logging/user.env` file. This variable modifies the deployment of Open Distro for Elasticsearch to be TLS-enabled.
+### Ingress
+
+* If you are using ingress, set the environment variable `TLS_ENABLED=true` in the `user.env` file in order to use TLS for connections between the user and Kibana.
```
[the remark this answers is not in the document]
gsmith-sas: @BryanEllington  that is (now) incorrect.  With the code change pushed as part of PR #57, setting TLS_ENABLED to "true" will require HTTPS b/w the browser and Kibana even if Ingress is not configured.

samples/tls/logging/README.md
```diff
@@ -37,9 +24,26 @@ my_repository_path/logging/bin/deploy_logging_open.sh
...
+you run the deployment script:
+
+  * kubernetes.io/tls secret - `kibana-ingress-tls-secret`
+  * kubernetes.io/tls secret - `elasticsearch-ingress-tls-secret`
+
+Generating these certificates is outside the scope of this example. However, you
+can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq) in SAS Viya Administration documentation and specify the `logging` namespace.
+
+If any of the required certificates do not exist, the deployment process attempts to use [cert-manager](https://cert-manager.io/) to generate the missing certificates.
+
```
gsmith-sas, line 41: Is this true?  I do not believe there is anything in the log deployment scripts that checks for these particular certificates or secrets.  I think the user is required to manually create these secrets (end of story).

samples/tls/logging/README.md
```diff
@@ -37,9 +24,26 @@ my_repository_path/logging/bin/deploy_logging_open.sh
...
+* If you are using an ingress controller other than NGINX, modify the annotation 
+`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` in the `user-values-elasticsearch-open.yaml` file. Refer to the documentation for your ingress controller.
```
gsmith-sas, line 43: Maybe add an "as needed into the sentence; i.e. "_modify the annotation...**as needed** in the user-values-elasticsearch-open.yaml file."_

samples/tls/logging/README.md
```diff
@@ -37,9 +24,26 @@ my_repository_path/logging/bin/deploy_logging_open.sh
...
+can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq) in SAS Viya Administration documentation and specify the `logging` namespace.
+
+If any of the required certificates do not exist, the deployment process attempts to use [cert-manager](https://cert-manager.io/) to generate the missing certificates.
+
+* If you are using an ingress controller other than NGINX, modify the annotation 
+`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` in the `user-values-elasticsearch-open.yaml` file. Refer to the documentation for your ingress controller. 
+
+### Nodeports
+
+If you are using nodeports,  set the environment variable `TLS_ENABLED=true` in the `user.env` file in order to use TLS for connections between the user and Kibana.
```
gsmith-sas, line 48: The Nodeports section can probably be removed if/when the sentence about setting TLS_ENABLED=true" if moved out of the Ingress section (since, as commented above, the setting has effect even if ingress is not enabled).  Hmmm, we need to look into whether the es_nodeport_enable.sh script needs to be tweaked to do something different if TLS_ENABLED is true (i.e. can we force HTTPS for inbound calls to Elasticsearch?).

samples/tls/monitoring/README.md
gsmith-sas: @BryanEllington are you saying browser to Grafana/Prometheus/AlertManager traffic requires HTTPS by default?  That's not what I've seen in my deployments (all connections are HTTP and shown as "Not secure" in my browser).

samples/tls/monitoring/README.md
```diff
@@ -64,17 +28,44 @@ my_repository_path/monitoring/bin/deploy_monitoring_cluster.sh
...
+
+Generating these certificates is outside the scope of this example. However,
+you can use the process documented in ["Configure NGINX Ingress TLS for SAS Applications"](https://go.documentation.sas.com/?cdcId=sasadmincdc&cdcVersion=default&docsetId=calencryptmotion&docsetTarget=n1xdqv1sezyrahn17erzcunxwix9.htm&locale=en#n0oo2yu8440vmzn19g6xhx4kfbrq)
+in SAS Viya Administration documentation and specify the `monitoring` namespace.
+
+* If you are using ingress and are using an ingress controller other than NGINX, modify the annotation 
+`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` in the `user-values-prom-operator.yaml` file. Refer to the documentation for your ingress controller. 
+
+### Secrets for In-Cluster TLS
+
```
gsmith-sas, line 52: We do NOT have and probably should add a similar section in the corresponding logging readme file.  The secrets are 

- es-rest-tls-secret
- es-transport-tls-secret
- kibana-tls-secret 

The logging deployment scripts will use cert-manager to generate these if the expected secrets are not found.

gsmith-sas (commented):
See review comments below.

samples/tls/logging/README.md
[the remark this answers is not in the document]
BryanEllington: I noted "over ingress" which remains true. You can set TLS_ENABLED=true and whether you configure ingress for TLS or not is a separate, orthogonal decision.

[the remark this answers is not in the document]
davidbass13: Changed this sentence to "If you are using ingress, connections between the user and logging components use TLS by default."

samples/tls/logging/README.md
davidbass13: Changed as noted

samples/tls/monitoring/README.md
davidbass13: Section added

samples/tls/monitoring/README.md
davidbass13: Intro sentences changed to "If you enable TLS, in-cluster communications between monitoring components use TLS. 

If you enable TLS and are using nodeports, connections between the user and monitoring components also use TLS."

samples/tls/monitoring/README.md
davidbass13: Section removed

samples/tls/logging/README.md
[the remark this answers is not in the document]
gsmith-sas: Sorry I misread Bryan's original comment which has only added to the confusion.  

1.  Regardless of whether you use ingress or not, the connections between the various logging components themselves use TLS.  
2. Communication between the user and Kibana (w/o ingress) do not use TLS by default but will if you set TLS_ENABLE (or LOG_KB_TLS_ENABLE) to true.  
3. Communication between the user and Elasticsearch (directly) is not enabled by default when using nodePorts.  However, if the es_enable_nodeport.sh script is run, making direct interaction with Elasticsearch available, TLS (HTTPS) will be used always.   
4. A user could choose to deploy/create Ingress objects to sit b/w the user and Kibana.  If they do that and set TLS_ENABLE (or LOG_KB_TLS_ENABLE) to true, communication between the ingress object they created and Kibana will use TLS.  
5. Similarly, a user could choose to deploy/create Ingress objects to sit b/w the user and (direct access to) Elasticsearch. If they do that, communication between that ingress object and Elasticsearch will always use TLS, regardless of the value of any env vars.

I think that covers all of the paths, and I hope clarifies things.

samples/tls/monitoring/README.md
gsmith-sas: That's true but just reflects the need for TLS to communicate with the monitoring objects rather than any "extra"/"additional" (configuration) work.  Just mentioning that as clarification, not necessarily as something that needs to go into the doc.

samples/tls/logging/README.md
BryanEllington: Yes, the user is required to provide any certificates referenced in the **ingress** section for both monitoring and logging.

samples/tls/logging/README.md
davidbass13: Section added to main logging README

samples/tls/logging/README.md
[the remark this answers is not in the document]
davidbass13: Statement added "All communication within the cluster between the logging components takes place through TLS-enabled connections."

logging/README.md
```diff
@@ -261,3 +261,13 @@ deployment process.
...
 __Add filter__ to filter the messages that are displayed.
+
+## Enable Users to Access Elasticsearch (optional)
+
+You can choose to enable users to access Elasticsearch so that they can issue queries using API calls or scripts. Run this command to enable nodeport connections to Elasticsearch:
+
+```bash
+./logging/bin/es_nodeport_enable.sh
+``` 
+
```
gsmith-sas, line 272: When the es_nodeport_enable.sh script completes, it writes a message with the URL you can use to access Elasticsearch (similar to the message w/the Kibana URL).  For example:

```
=============================================
== Access Elasticsearch using this URL: https://myK8snode:32168/ ==
=============================================
```

However, unlike in the Kibana message, the port number is not fixed (i.e. not always 32168) and users would need to use the port number shown in the message.  Notice the protocol is https (indicating the TLS is enabled) and that this is always the case.
Probably should mention that there is a corresponding script to disable the nodeport enabled by running this script.  Unimaginatively, the disable script is /logging/bin/es_nodeport_disable.sh.  We might also want to mention that these scripts can be useful if users need to temporarily enable access for some reason.

samples/tls/logging/README.md
gsmith-sas: The **kubectl create secret** command does NOT create TLS certs; it creates Kubernetes secret objects.  Once you have created a TLS cert through some mechanism, you can store them within a Kubernetes secret object to make them available to the deployment scripts and Kubernetes pods.

Do we want to tell people to use cert-manager to create the certs?  Or, should we just tell them they need to create/obtain the TLS certs and then store them in Kubernetes secret objects?  If we mention the kubectl command, we probably need to provide a more complete example of the syntax.  We have some syntax shown in the NOTES_ON_TLS.md file although those are specific to what we need for the Elasticsearch TLS certs -- not sure if the ingress cert secrets are the same.

samples/tls/logging/README.md
```diff
@@ -37,9 +26,36 @@ my_repository_path/logging/bin/deploy_logging_open.sh
...
+Use cert-manager, if available, to create these certificates. If cert-manager is not available, you must create the certificates manually by using the `kubectl -n logging create secret` command.
+
+* If you are using an ingress controller other than NGINX, modify the annotation 
+`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` as needed in the `user-values-elasticsearch-open.yaml` file. Refer to the documentation for your ingress controller. 
+
+### Nodeports
+
+* If you are using nodeports, set the environment variable `TLS_ENABLED=true` in the `user.env` file to use TLS for connections between the user and Kibana.
+
+* If users need to directly connect to Elasticsearch, you must run the `es_enable_nodeport.sh` script to enable the connection. Direct connections to Elasticsearch always use TLS, regardless of the value of the `TLS_ENABLED` variable.
```
gsmith-sas, line 48: Perhaps: _"If users need to directly connect to Elasticsearch **via a nodePort**, you must run the..."_  

Hmmm, we now cover this in the top-level logging README file and it doesn't really have anything to do with TLS, so maybe remove this paragraph completely.

samples/tls/logging/README.md
gsmith-sas: After giving this a bit more thought, how about doing the following:

- Add a new section (under Customization) called TLS, and explain how to set the TLS_ENABLE env var to "true" to enable TLS (in the one spot it isn't being used).
- The Ingress section should only talk about the need for those annotations and the need to populate the secrets with the actual TLS certs.
- Remove the Nodeport section.

I think that's cleaner and would make more sense to the reader.

samples/tls/monitoring/README.md
```diff
@@ -64,20 +27,38 @@ my_repository_path/monitoring/bin/deploy_monitoring_cluster.sh
...
 any sample hostnames with hostnames for your deployment. Specifically, you must replace
 `host.cluster.example.com` with the name of the ingress node. Often, the ingress node is the cluster master node, but your environment might be different.
 
-## Limitations and Known Issues
+* If you are using ingress, manually populate these secrets in the `monitoring` namespace (or `MON_NS` value) **before** you deploy cluster monitoring.
+
+* kubernetes.io/tls secret - `prometheus-ingress-tls-secret`
+* kubernetes.io/tls secret - `alertmanager-ingress-tls-secret`
+* kubernetes.io/tls secret - `grafana-ingress-tls-secret`
+
```
gsmith-sas, line 41: The notation here (and in the corresponding section in the logging readme file) is probably not clear to the reader.  The first part (e.g. kubernetes.io/tls secret) is just identifying the type of Kubernetes object to create.  The second part (alertmanager-ingress-tls-secret) identifies the name of the Kubernetes secret object within the monitoring (or logging) namespace that contains the TLS cert.    I think some sample syntax (i.e. syntax needed to create one of these secrets from a TLS cert file) would be helpful here.  @BryanEllington can you provide David with that?

gsmith-sas (commented):
I noticed that, in at least one place, we referred to the env var as TLS_ENABLED, this is incorrect.  The env var is TLS_ENABLE.

See other comments/suggestions in-line below.

samples/tls/logging/README.md
davidbass13: That does make the organization cleaner. Changed as noted.

logging/README.md
davidbass13: Message and note about disable script added

samples/tls/logging/README.md
davidbass13: Agreed - removed

samples/tls/logging/README.md
BryanEllington: I don't think we should get into how to create the various .crt files - there's CAs and server certs and there are nuances and caveats for each. It's out of scope for this project, imo.

gsmith-sas: I agree that we shouldn't get into the weeds of how to create certs.  But, assuming they have a set of certs, what's the syntax for creating the corresponding Kubernetes secrets?    For the Elasticsearch TLS certs, we don't use Kubernetes TLS secret objects, just Kubernetes generic secrets.  But, I believe the secrets for the ingress certs must be Kubernetes TLS secrets; correct?  If so, is this the syntax (taken from kubectl --help text)?
`kubectl create secret tls NAME --cert=path/to/cert/file --key=path/to/key/file`

BryanEllington: `kubectl create secret tls "$SECRET_NAME" -n "$NAMESPACE" --key "$CERT_KEY" --cert "$CERT_FILE"`

logging/README.md
gsmith-sas: Thanks.  One small nit:  The formatting of the Elasticsearch message and the Kibana message shown elsewhere in the doc are not consistent.  While the port number is variable here (compared to the Kibana case where it is hard-coded) and, may, therefore, warrant different treatment,  the way the hostname is shown differs b/w the two examples too.  I have a slight preference for the way things are shown w/the Kibana message but am flexible and am more concerned about consistency.

samples/tls/logging/README.md
```diff
@@ -37,9 +26,37 @@ my_repository_path/logging/bin/deploy_logging_open.sh
 ```
 ## Notes On Customization Values
 
-Set `MON_TLS_ENABLE=true` in the `$USER_DIR/logging/user.env` file. This variable modifies the deployment of Open Distro for Elasticsearch to be TLS-enabled.
+### TLS 
+
+Specify `TLS_ENABLE=true` in the `user.env` file to use TLS for communications between the ingress object and Kibana. Connections between the ingress object and Elasticsearch always use TLS, regardless of the value of `TLS_ENABLE`.
```
gsmith-sas, line 31: Thanks for moving this sentence out...just need to tweak the wording to remove the references to ingress objects (since this flag has an impact, even if ingress is not used).  Perhaps:
_Specify...to require TLS for connections to Kibana.  Connections to Elasticsearch itself (when enabled) always requires TLS, regardless of the value of TLS_ENABLE._

samples/tls/monitoring/README.md
```diff
@@ -64,24 +27,49 @@ my_repository_path/monitoring/bin/deploy_monitoring_cluster.sh
...
-Grafana, and Alertmanager to be TLS-enabled.
+* Set `TLS_ENABLE=true` in the `$USER_DIR/monitoring/user.env` file to specify that Prometheus, Grafana, and Alertmanager are deployed with TLS enabled for connections to these components. Due to limitations in the underlying Helm charts, some components might not have TLS enabled for in-cluster connections between components. See the **Limitations and Known Issues** section below for details.
 
-Edit `$USER_DIR/monitoring/user-values-prom-operator.yaml` and replace
+* Edit `$USER_DIR/monitoring/user-values-prom-operator.yaml` and replace
 any sample hostnames with hostnames for your deployment. Specifically, you must replace
 `host.cluster.example.com` with the name of the ingress node. Often, the ingress node is the cluster master node, but your environment might be different.
 
-## Limitations and Known Issues
+* Manually populate these TLS secrets for ingress in the `monitoring` namespace (or `MON_NS` value) **before** you deploy cluster monitoring.
```
gsmith-sas, line 36: Seems like this item is only relevant if the reader wants to deploy ingress objects, so we probably should call that out.  This sentence and the one that follows don't flow.
Perhaps something like:
_If you plan on using ingress, before you deploy cluster monitoring, you must manually populate specific Kubernetes secrets with TLS certificates.  The following Kubernetes secrets must be populated:
...
You can use the following command to generate the necessary secret for each set of TLS certificates:_

gsmith-sas (commented):
Looking good...but still needs a few small tweaks.

samples/tls/logging/README.md
davidbass13: Changed as noted

samples/tls/monitoring/README.md
davidbass13: Changed as noted

logging/README.md
davidbass13: Changed message format and wording to be more consistent with Kibana example

gsmith-sas (commented):
All of these changes look good.  I took another look at the whole samples/tls/logging/README.md file and noticed a couple of small things:

- The OVERVIEW section is now 3 sentences formatted as 3 paragraphs.  I think they should probably combined into one paragraph.  Some tweaking of the wording may be needed; perhaps:
   - _Communication between the logging components within the cluster always takes place through TLS-enabled connections.  This sample demonstrates how to deploy enable TLS for connections between the user (and/or an ingress object) and Kibana. The TLS_ENABLE environment variable controls whether connections with Kibana use TLS or not._
- Probably should mention and include a pointer to the NOTES_ON_USING_TLS.md file (in the logging sub-dir of the repo) in the section Secrets for In-Cluster TLS

Other than those suggestions, I think this looks good to go.
- Last sentence mention that cert-manager is not required if TLS is not enabled.  Since TLS is always enabled for logging, this is incorrect.  But the other condition cited (i.e. certs already exist) is still a valid and the sentence should remain.  It would be nice to find a way to avoid starting the sentence with the (lower-cased) word cert-manager...maybe _The cert-manager component|applications is not required if..._

BryanEllington (approved):
Looks good to me

gsmith-sas (approved):
Looks good to me.

PR #60, opened 2021-01-26
added ignore case for node to the kube node dashboards

Vladislaff:
In an azure aks cluster the name of the node differs with the hostname by one last letter - it has a different case.
Example:
kubectl get nodes returns aks-cas-23491758-vmss00000q
hostname from the node (and the instance name in the vmss in azure) is aks-cas-23491758-vmss00000Q

Because of this the exported_nodename and nodename are different and the kube node dashboards do not show metrics for the instances which end in a letter.


BryanEllington (approved):
Verified it works outside of Azure. Looks good.

PR #62, opened 2021-01-28
Move to fluent/fluent-bit Helm chart from stable/fluent-bit chart

gsmith-sas:


logging/bin/deploy_fluentbit_azmonitor.sh
```diff
@@ -73,6 +75,13 @@ else
    export AZMONITOR_SHARED_KEY=$(kubectl -n $LOG_NS get secret connection-info-azmonitor -o=jsonpath="{.data.shared_key}" |base64 --decode)
 fi
 
+# Check for an existing Helm release of stable/fluent-bit
+if helm3ReleaseExists fbaz $LOG_NS; then
+  log_info "Removing an existing release of deprecated stable/fluent-bit Helm chart from from the [$LOG_NS] namespace [$(date)]"
+  helm  $helmDebug  delete -n $LOG_NS fbaz
```
gsmith-sas, line 81: One potential issue: If we are upgrading from a previous deployment that used the stable Helm repo, user will need to (re-)run monitoring/bin/deploy_monitoring_cluster.sh to get the new serviceMonitor for (new version of) Fluent Bit.  Should we display a message telling them that?

BryanEllington: Good point. I would say you could check for the 'v4m-prometheus-operator' release, but you really wouldn't know the namespace. I'd hate to log a warning if they don't even have monitoring installed. Maybe an info message, maybe just doc in the CHANGELOG?

BryanEllington (approved):
LGTM

PR #74, opened 2021-02-17
Support additional Grafana dashboards in USER_DIR

BryanEllington:


monitoring/README.md
```diff
@@ -175,6 +175,34 @@ the following jobs in the Istio Prometheus instance:
 * `kubernetes-pods`
 * `kubernetes-service-endpoints`
 
+### Add Custom Dashboards
```
BryanEllington, line 178: This section describes two of the three ways to deploy additional dashboards. We should probably mention the `$USER_DIR/monitoring/dashboards' method here in addition to the section in the generic-base sample dashboards directory.

monitoring/README.md
```diff
@@ -175,6 +175,39 @@ the following jobs in the Istio Prometheus instance:
...
+
+#### Manually Import Dashboards
+
+You can use Grafana's import function to import dashboards. 
+
+The advantage of using the import function is that Grafana recognizes if the dashboard's data source is not valid for your environment and prompts you to select your environment's data source. 
+
+The disadvantage is that the process is not automated, so you must re-import the dashboards each time you redeploy the monitoring components. If you have many dashboards to import, the process could also be time-consuming. 
+
+### Add Dashboards Using a Script
```
gsmith-sas, line 195: @davidbass13 I think this needs to be a level 4 header (to match Manually Import Dashboards).

monitoring/README.md
```diff
@@ -175,6 +175,39 @@ the following jobs in the Istio Prometheus instance:
...
+
+You can use the `deploy_dashboards.sh [dashboard | dashboard_directory]` script to deploy a single dashboard or a directory of dashboards.
+
+Grafana cannot validate the data source of dashboards that are deployed using this script, so you must manually change the data source to `Prometheus` in the dashboards' JSON files before you deploy them. For example:
+- Find: `"datasource": "${DS_PROMETHEUS}"`
+- Replace: `"datasource": "Prometheus"`
+
```
gsmith-sas, line 202: Change "Replace" with "Replace with"?

monitoring/README.md
```diff
@@ -175,6 +175,39 @@ the following jobs in the Istio Prometheus instance:
...
+The advantage of using the script is that the process can be automated, so you can automatically re-deploy all of your custom dashboards whenever you re-deploy the monitoring components.
+
+Some detailed dashboards might be too large to deploy using the script, and must be imported manually. 
+
```
gsmith-sas, line 206: What happens if the dashboard is too large?  Does it fail with a helpful message?  Die mysteriously?

monitoring/README.md
```diff
@@ -175,6 +175,39 @@ the following jobs in the Istio Prometheus instance:
...
+### Deploy Dashboards in the $USER_DIR/monitoring/dashboards Directory
+
+Dashboards that are stored in the `$USER_DIR/monitoring/dashboards` directory are automatically added when you deploy the monitoring components. The dashboards must be in `.json` format, and they cannot be in a subdirectory under `$USER_DIR/monitoring/dashboards`. If you remove the monitoring components using the `remove_monitoring_cluster.sh` script, the dashboards are removed as well.
+
```
gsmith-sas, line 210: Do the issues related to datasources still exist with this option?  If so, we should mention that.  We probably also want to highlight this option is best used _after_ fixing dashboards imported using the other techniques and then exporting them so they can be deployed repeatedly in the future.

gsmith-sas (commented):
Just a few comments mostly on the doc

monitoring/README.md
davidbass13: Agreed - changed

monitoring/README.md
davidbass13: Changed

monitoring/README.md
davidbass13: Added note that a message is displayed

monitoring/README.md
davidbass13: Sectino added

gsmith-sas (commented):
Looks like David's revisions have addressed my documentation concerns.

gsmith-sas (approved):

PR #75, opened 2021-02-17
Udpate logging certificate custom resources to cert-manager v1

BryanEllington:


gsmith-sas (approved):
I've confirmed the certs match our expected structure.

PR #82, opened 2021-03-09
Generate application access URLs from available information (inc. ingress)

gsmith-sas:


logging/bin/deploy_kibana_content.sh
```diff
@@ -167,15 +168,44 @@ rm -f $tmpfile
 sleep 7s
 
 
+set +e
+# call function to get HTTP/HTTPS ports from ingress controller
+get_ingress_ports
+
+# get URLs for Kibana and Elasticsearch REST api endpoint
+kb_url=$(get_service_url logging v4m-es-kibana-svc "/app/kibana" "$LOG_KB_TLS_ENABLE" "v4m-es-kibana")
```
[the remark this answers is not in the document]
BryanEllington: The app/kibana path is unnesessary, so yes, get rid of it if you can

logging/bin/service-url-include.sh
```diff
@@ -0,0 +1,193 @@
...
+#    * The get_service_url function assumes variables ingress_http_port and ingress_https_port have been set
+#    * call  get_ingress_ports function to set these variables or set them by hand
+#
+# # Sample usage:
+#   grafana_url=$(get_service_url monitoring v4m-grafana  "/" "false")
+#
+# Returns generated URL or "" (null) string (if unable to generate URL)
+# Exit code: 0 = success 1 = failure to generate URL
+#
+
```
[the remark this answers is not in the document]
BryanEllington: I do think we need to document all of our functions better. I'd like to pick a standard header block and apply that everywhere. That can be for another day, though. As for overall developer usage, I can see that being useful, but a lower priority than the function documentation.

logging/README.md
```diff
@@ -239,23 +239,13 @@ The script removes configmaps and secrets that were created by the deployment sc
 
 ### Access Kibana
 
-If the deployment process completes without errors, a message such as this
-appears in the console window:
+If the deployment process completes without errors, a message 
+appears in the console window containing the URL address for Kibana and the URL for Elasticsearch (if you enabled access to Elasticsearch). The URL for Kibana is different depending on whether you use nodeports or ingress to access Kibana.
```
gsmith-sas, line 243: The URL is different depending on how things are configured, it isn't just that there's one URL if nodeports are in use and another if ingress is configured.  So, I think the last sentence in this paragraph can be dropped or it needs to be reworded.  

Also, we probably should clarify that the URL shown for Elasticsearch is really only useful for API access and is not designed to used in a web browser.  Maybe we should refer to it as the "URL for the Elasticsearch API endpoint"?

logging/README.md
```diff
@@ -239,23 +239,13 @@ The script removes configmaps and secrets that were created by the deployment sc
...
-The message provides the URL address for the Kibana application. To validate
-that the deployment was successful and confirm that all of the logging components
+To validate that the deployment was successful and confirm that all of the logging components
 are working, access Kibana and review the log messages that are collected.
 
-__Note:__ The displayed URL for Kibana might not be correct if you defined
-ingress objects or if your networking rules alter how hosts are accessed. If
-this is the case, contact your Kubernetes administrator to determine the proper
-host, port and/or path to access Kibana.
+__Note:__ The displayed URL for Kibana might not be correct if your networking rules alter how hosts are accessed. If this is the case, contact your Kubernetes administrator to determine the proper host, port and/or path to access Kibana.
```
gsmith-sas, line 248: No need to limit this to Kubernetes; perhaps: _The URLs displayed might not be correct..._

gsmith-sas (commented):
Just a few comments...

logging/README.md
BryanEllington: Or just don't display the Elasticsearch URL. It's not necessary in most cases and our output should be reserved for only the most important services - Grafana and Kibana.

BryanEllington (approved):
LGTM

PR #91, opened 2021-03-22
Added note about deployment in highly-secure environments

davidbass13:
Added a note in the pre-deployment section about task required when deploying in a highly-secured environment.

### DIFF
--- a/logging/README.md
+++ b/logging/README.md
@@ -32,7 +32,7 @@ Before deploying, you must perform these tasks:
 - [create a local copy of the repository](#log_loc_copy)
 - [customize your deployment](#log_custom)
 
-Note: If you are deploying in a highly secure environment (in which pods cannot connect to the internet and network traffic is routed through a proxy server), ensure that `localhost` is included in the list of hosts that are not routed through the proxy server. If it is not included, the Elasticsearch deployment fails, because the deployment scripts use port forwarding to make REST API calls to Elasticsearch and Kibana.  
+Note: If you are deploying in a highly secure environment in which the deployment scripts are run on a machine whose network traffic is routed through a proxy server, ensure that `localhost` is included in the list of hosts that are not routed through the proxy server. If it is not included, the Elasticsearch deployment fails, because the deployment scripts use port forwarding to make REST API calls to Elasticsearch and Kibana.  
 
 ### <a name="log_loc_copy"></a>Create a Local Copy of the Repository
 
--- a/logging/README.md
+++ b/logging/README.md
@@ -32,6 +32,8 @@ Before deploying, you must perform these tasks:
 - [create a local copy of the repository](#log_loc_copy)
 - [customize your deployment](#log_custom)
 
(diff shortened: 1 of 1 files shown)


logging/README.md
```diff
@@ -32,6 +32,8 @@ Before deploying, you must perform these tasks:
 - [create a local copy of the repository](#log_loc_copy)
 - [customize your deployment](#log_custom)
 
+Note: If you are deploying in a highly secure environment (in which pods cannot connect to the internet and network traffic is routed through a proxy server), you must add `localhost` to the list of hosts that are not routed through the proxy server. If you do not, the Elasticsearch deployment fails.
```
gsmith-sas, line 35: I'm not sure the pod connectivity issue is relevant in this particular case.  I think the use of the proxy server is the important restriction.  And, to be honest, although I could be wrong about this, I suspect most of the time "localhost" has already been added to the proxy list.  I think this particular customer just happened to have omitted it.  So, I wonder if we should phrase it as "_...be sure 'localhost' has been added to the lists..._" rather than telling them to do it.  But I may be reading more into the wording than is there.  If you want to provide more context/explanation, you could add that the deployment scripts use port forwarding to make REST API calls to Elasticsearch and Kibana as the scripts execute.

gsmith-sas (changes requested):

logging/README.md
davidbass13: Revised as noted.

logging/README.md
```diff
@@ -32,7 +32,7 @@ Before deploying, you must perform these tasks:
 - [create a local copy of the repository](#log_loc_copy)
 - [customize your deployment](#log_custom)
 
-Note: If you are deploying in a highly secure environment (in which pods cannot connect to the internet and network traffic is routed through a proxy server), you must add `localhost` to the list of hosts that are not routed through the proxy server. If you do not, the Elasticsearch deployment fails.  
+Note: If you are deploying in a highly secure environment (in which pods cannot connect to the internet and network traffic is routed through a proxy server), ensure that `localhost` is included in the list of hosts that are not routed through the proxy server. If it is not included, the Elasticsearch deployment fails, because the deployment scripts use port forwarding to make REST API calls to Elasticsearch and Kibana.
```
gsmith-sas, line 35: I like the revisions.  But, thinking about this further I realized pod connectivity is really not relevant here.  The issue we're addressing relates to the network configuration of the machine on which the deployment script is running rather than any of the pods.  In fact, the deployment scripts could be (likely are) run on a machine outside of the K8s cluster completely.  Therefore, the first sentence should be revised to clarify that if the deployment scripts are being run on a machine which has its network traffic routed through a proxy server, the proxy settings need to include "localhost" in the list of "acceptable" hosts.

gsmith-sas (commented):
I like the revisions, but see my note about clarifying where the proxy server is coming into play.

gsmith-sas (approved):
Looks good.  Thanks!

PR #93, opened 2021-03-30
Reorganization of customization information in main README

davidbass13:


logging/README.md
```diff
@@ -59,13 +61,40 @@ in a directory named `viya4-monitoring-kubernetes-<release_number>`.
...
 - using the `USER_DIR` environment variable to specify the location of the customization files
 - copying the customization files from one of the provided samples to your local directory
 - specifying customization variables and parameters in the customization files
 
-See the [main README](../README.md#customization) to for information about the customization process.
+Other scenarios use different customization steps that are specific to each scenario.
+
+[Samples](#lsample) are provided for several common deployment scenarios. Each sample includes detailed information about the customization process and values for the scenario. 
+
+#### Creating the Location for Customization Files
```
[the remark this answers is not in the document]
davidbass13: The paragraph about USER_DIR has this statement - "By using a location outside of the repository, you can put the customization files in a location that is under version control, and you can preserve your customizations when you deploy new versions of the monitoring and logging applications." Is there more that's needed, or does it need to say something different?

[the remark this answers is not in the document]
BryanEllington: Missed that. Cool.

BryanEllington (approved):

monitoring/README.md
```diff
@@ -66,13 +66,40 @@ repository is created in a directory named `viya4-monitoring-kubernetes-<release
...
+
+#### Creating the Location for Customization Files
+
+The `USER_DIR` environment variable enables you to use a directory outside of the local repository to contain the customization files. By using a location outside of the repository, you can put the customization files in a location that is under version control, and you can preserve your customizations when you deploy new versions of the monitoring and logging applications. You could also create different sets of configuration files for different cluster types, and then use the `USER_DIR` variable to specify the configuration files to use based on the cluster type on which you were deploying.
+
+For example, to create a customization directory named *my-viya4mon-user-dir* that contains a directory for monitoring customization files:
+
+```bash
+mkdir -p ~/my-viya4mon-user-dir/monitoring
+```
```
gsmith-sas, line 87: I wonder if it is confusing to readers to create this multi-level directory (i.e. all the way down to the monitoring sub-dir) and then, as directed in the next section, point the USER_DIR env var one level higher.  We could: 

-  break the make directory step into two steps (i.e. create my-viya4mon-user-dir and then create the monitoring sub-dir under it);
- or, maybe just point out that the env var is pointing to the parent directory?

davidbass13: I've revised the instructions for creating the directory to include two steps - creating the root and then the sub-directory and indicated that the root directory is where USER_DIR will point

gsmith-sas: Looks good.  You can remove the -p option in the mkdir commands.  It's a flag that says "make any parent/intermediate directories that don't already exist" and no longer needed since we're now creating each level manually.

gsmith-sas (commented):
Looks good...just one comment about the syntax of the mkdir commands.

monitoring/README.md
davidbass13: Great - thanks. Fixed as noted

PR #94, opened 2021-04-01
Improve cloudwatch Prometheus metric mapping

BryanEllington:


davidbass13:
Long entries in the far right-hand column of all three tables in reference.md are cut off.

samples/cloudwatch/README.md
```diff
@@ -92,25 +122,59 @@ kubectl delete pod -n amazon-cloudwatch -l app=cwagent-prometheus
...
+
+After you find the metric you are interested in, you can create a graph from
+the metric data. For example, these are the steps to create a graph of
+`go_memstats_alloc_bytes` (memory usage) of SAS Viya services that are
+written in Go:
+
+1. Go to [CloudWatch SAS Viya Metrics](reference.md) and expand the
+**By Metric** table.
+2. Locate the `go_memstats_alloc_bytes` metric. The table indicates that it is
+associated with the `container,job,namespace,node,pod,sas_service_base,service`
```
BryanEllington, line 159: It's under 'ClusterName,job,namespace,node,pod,sas_service_base,service'

samples/cloudwatch/README.md
```diff
@@ -92,25 +122,59 @@ kubectl delete pod -n amazon-cloudwatch -l app=cwagent-prometheus
...
+dimensions.
+3. From the AWS web console, navigate to CloudWatch.
+4. In the navigation bar on the left, choose `Metrics`.
+5. Select the `ContainerInsights/Prometheus` namespace.
+6. Select the tile with the `container,job,namespace,node,pod,sas_service_base,service`
```
BryanEllington, line 164: ClusterName,job,namespace,node,pod,sas_service_base,service

samples/cloudwatch/README.md
davidbass13: fixed

samples/cloudwatch/README.md
davidbass13: fixed

davidbass13:
I think it might improve the usefulness of the tables if the table headers (dimension, source, or metric) were formatted using the /details tag, so they could also be collapsible. Some of those tables are really long, so rather than making users scroll endlessly, they could quickly locate what they were looking for and expand the table they need.

The Dimensions column also looks narrow in the By Source table, especially since there's so much white space. I don't know if that's a result of keeping the text from getting cut off, though. If it has to stay like that, it's not the worst thing.

BryanEllington:
I agree, and I tried, but couldn't get nested collapsible sections working correctly.

davidbass13:
Trying things out in VS Code and looking at it in the Preview window, it looks like this coding works:
<details>
<summary>ClusterName,cas_server,job,namespace,pod</summary>
table...
</details>

davidbass13:
Sorry...the comment rendered my code.  The coding is to put details before the dimension/source/metric, wrap the dimension/source/metric in summary, and put an end details tag after the table.

BryanEllington:
Yes, that works by itself - but not nested tables inside another details section.

BryanEllington:
OK...I must have been doing something wrong. Seems to work now. I don't know how to do an 'expand all', so maybe it will be annoying to have to expand each section. Take a look at the latest commit.

davidbass13:
I think that's much better. Even though you can't do an "expand all", I think that's preferable to having to do all that scrolling.

samples/cloudwatch/reference.md
```diff
@@ -0,0 +1,2926 @@
...
+| cas_node_inodes_used | sas-cas |
+| cas_node_max_open_files | sas-cas |
+| cas_node_one_minute_cpu_load_avg | sas-cas |
+| cas_node_open_files | sas-cas |
+
+</details>
+<hr/>
+<details>
+  <summary>ClusterName,cas_node,cas_node_type,cas_server,connected,job,namespace,pod,uuid (1)</summary>
+
```
gsmith-sas, line 26: It would be nice to use different formatting and/or add text (maybe _**Dimensions:**_) to make the headers standout (or appear different) than the "Click to expand" text.  Once you expand, you see additional expandable sections but what the headers are confusion unless you know what they are.

samples/cloudwatch/README.md
```diff
@@ -2,52 +2,75 @@
...
-- The default CloudWatch agent collects metrics from AWS nodes. Although these metrics are not specific to SAS Viya, you might want to use both agents so you can obtain a complete view of your environment's performance. For information about using the default agent, see the 
-[CloudWatch documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Install-CloudWatch-Agent.html). If you do use the default agent, you must add tolerations to the
-agent's daemonset definition so that the agent can run on nodes that are tainted to support SAS Viya workload node placement. See [CloudWatch Agent Daemonset and Workload Node Placement](#wnp_tolerations).
+AWS uses the CloudWatch agent to collect metrics. There are two types of
+CloudWatch agents that you might want to use in your environment:
+
+- The CloudWatch agent with Prometheus monitoring scrapes metrics from
+Prometheus sources, such as SAS Viya. The metrics can then be converted and
+mapped for display in CloudWatch. Because SAS Viya monitoring uses
+Prometheus, this sample explains how to deploy and configure only the
```
gsmith-sas, line 19: I think this last sentence needs to be more prominent or broken out as its own paragraph; it is too easily missed otherwise.

samples/cloudwatch/README.md
```diff
@@ -2,52 +2,75 @@
...
 
 ### Identify Cluster and Region
 
-You must identify the Amazon EKS cluster on which SAS Viya is deployed and the region for the cluster. These values are needed to construct the correct endpoint for CloudWatch.
+You must identify the Amazon EKS cluster on which SAS Viya is deployed and the
+region for the cluster. These values are needed to construct the correct
+endpoint for CloudWatch.
 
 Run the following code from the command line. Replace the
 values of `ClusterName` and `RegionName` with the values for your EKS cluster.
```
gsmith-sas, line 76: Rather than saying "run the following code", maybe say _"Set the ClusterName and RegionName environment variables with the values for your EKS cluster"_  While setting env vars may, technically, be "submitting code", that seems to overstate things.
And are we only setting env vars so users can cut-and-paste the curl command shown afterwards?  Or, are the env vars needed/used later?  If it is the former, I think we should point that out (e.g. "having set the environment variables appropriately for your cluster, you can cut-and-paste the following command to...").

samples/cloudwatch/README.md
```diff
@@ -92,25 +122,59 @@ kubectl delete pod -n amazon-cloudwatch -l app=cwagent-prometheus
...
+graph.
 
 ## CloudWatch Agent Daemonset and Workload Node Placement<a name=wnp_tolerations></a>
 
 The SAS Viya workload node placement configuration uses annotations and node taints
 to schedule pods on appropriate nodes. The default CloudWatch agent is deployed
-as a daemonset and runs on every node. If you use the default CloudWatch agent, you must add tolerations to the agent's daemonset definition to enable the agent to run on nodes that are tainted for SAS Viya workload node placement.
+as a daemonset and runs on every node. If you use the default CloudWatch agent,
+you must add tolerations to the agent's daemonset definition to enable the agent
+to run on nodes that are tainted for SAS Viya workload node placement.
```
gsmith-sas, line 178: Do we need to tell them where to find or how to add the code snippet to the daemonset defintion?  We've provided very detailed instructions most every else, so this section seems very terse in comparison.

samples/cloudwatch/reference.md
```diff
@@ -0,0 +1,2926 @@
...
+</details>
+<hr/>
+<details>
+  <summary>ClusterName,cas_server,job,namespace,pod,state (1)</summary>
+
+| Metric | Source |
+| ------ | ------ |
+| cas_grid_state | sas-cas |
+
```
gsmith-sas, line 53: Are the source values (e.g. sas-cas, sas-go, etc.) meaningful outside of this document?  Do they correspond to specific pods or something?  If not why include the hypen?  And is the "sas" prefix needed/helpful?

samples/cloudwatch/README.md
davidbass13: I moved the last sentence to a new paragraph as mentioned

samples/cloudwatch/README.md
davidbass13: Wording changed as suggested

samples/cloudwatch/README.md
davidbass13: I'm not sure I understand the question...the snippet is just below the text.

samples/cloudwatch/README.md
gsmith-sas: Thanks, I think that's more clear.  I think it might be even clearer if we move it to after the 2nd bullet point and explicitly state that this sample focuses on the first agent and not the second.  

Something like:
_Because SAS Viya monitoring uses Prometheus, this sample explains only how to deploy and configure the CloudWatch agent for use with Prometheus and does not discuss the default CloudWatch agent in great detail._

samples/cloudwatch/README.md
gsmith-sas: I guess I'm wondering if we need to tell them where to find the agent's daemonset definition and how to edit it.

samples/cloudwatch/README.md
davidbass13: Done

samples/cloudwatch/README.md
davidbass13: Added a link to the Cloudwatch doc that provides more information about the DaemonSet

samples/cloudwatch/reference.md
BryanEllington: Changed 'Click to expand' to the name of the field - 'Dimensions', 'Sources', ...

gsmith-sas (approved):
LGTM

PR #108, opened 2021-05-19
Log Monitoring on OpenShift

gsmith-sas:


logging/README.md
```diff
@@ -248,6 +251,8 @@ which you cloned the repository and issue this command:
 
 The script creates the namespace into which the components are deployed. By default, the components are deployed into the namespace `logging`.
 
+**Note:* If you are deploying SAS Viya Logging on OpenShift, you must follow the deployment process documented in [SAS Viya Logging on OpenShift](/openshift/openshift.md).**
```
BryanEllington, line 254: Link is broken - should be `/logging/openshift/openshift.md` or `openshift/openshift.md`

logging/bin/deploy_elasticsearch_content_open_openshift.sh
```diff
@@ -0,0 +1,225 @@
+#! /bin/bash
```
BryanEllington, line 1: What is different about this file vs. the non-OpenShift version? Do we need this much duplication?

logging/bin/deploy_esexporter.sh
```diff
@@ -66,6 +66,18 @@ else
   wnpValuesFile="$TMP_DIR/empty.yaml"
 fi
 
+# Deploying on OpenShift?
+OPENSHIFT_CLUSTER=${OPENSHIFT_CLUSTER:-false}
```
BryanEllington, line 70: I believe `OPENSHIFT_CLUSTER` should always be set now, so maybe this isn't needed

logging/bin/deploy_logging_open_openshift.sh
```diff
@@ -0,0 +1,140 @@
+#! /bin/bash
```
BryanEllington, line 1: Lots of duplication. Any way we can avoid the large number of new scripts that are slight variants of the core files?

logging/README.md
gsmith-sas: Moved the logging/openshift/openshift.md to logging/OpenShift.md to be consistent with monitoring; fixed references in top-level README.md.

logging/bin/deploy_elasticsearch_content_open_openshift.sh
gsmith-sas: The difference is this one deploys a 3rd log management policy object.  We _could_ make that conditional logic within the original script but we've, so far, kept the OpenShift-specific logic separate.  I lean towards keeping the separation for now but as OpenShift support becomes less experimental, we can look at merging this logic into the original script.

logging/bin/deploy_esexporter.sh
gsmith-sas: removed the comment and line.

logging/bin/deploy_logging_open_openshift.sh
gsmith-sas: See note above re:keeping OpenShift-specific in separate scripts for now (as we did on the monitoring side).  For this _particular_ script, given the differences in deployment steps b/w OpenShift and other cloud providers, I'm not sure consolidating into a single script makes sense.

bin/service-url-include.sh
```diff
@@ -23,10 +26,10 @@ function get_k8s_info {
   object=$2
   info_item=$3
 
-  info=$(kubectl -n $namespace get $object  -o=jsonpath=${json_paths[$info_item]})
+  info=$(kubectl -n $namespace get $object  -o=jsonpath=${json_paths[$info_item]} 2>/dev/null)
   rc=$?
 
-  if [ "$rc" == "0" ]; then
+  if [ ! -z  "$info" ]; then
```
BryanEllington, line 32: minor nit that you don't even have to fix, but you can use `-n` in place of `! -z`

logging/bin/remove_logging_open_openshift.sh
```diff
@@ -0,0 +1,41 @@
...
+# remove ServiceMonitors
+export DEPLOY_SERVICEMONITORS=${DEPLOY_SERVICEMONITORS:-true}
+logging/bin/remove_servicemonitors_open_openshift.sh
+
+#
+# remove logging components
+#
+# override-openshift check
+export CHECK_OPENSHIFT_CLUSTER=false
+logging/bin/remove_logging_open.sh
```
BryanEllington, line 35: Another nit...Instead of exporting, when you want to set a variable just to execute a specific script, just prefix it:
```bash
DEPLOY_SERVICEMONITORS=${DEPLOY_SERVICEMONITORS:-true} logging/bin/remove_servicemonitors_open_openshift.sh```

BryanEllington (approved):
Just a couple of nits. No required changes.

PR #109, opened 2021-05-20
Support OpenShift monitoring deployment/removal with scripts

BryanEllington:


monitoring/bin/remove_monitoring_openshift.sh
```diff
@@ -0,0 +1,100 @@
...
+
+log_info "Removing CA bundle..."
+kubectl delete --ignore-not-found cm grafana-trusted-ca-bundle
+
+log_info "Removing dashboards..."
+monitoring/bin/remove_dashboards.sh
+
+log_info "Removing service monitors..."
+kubectl delete --ignore-not-found podmonitor -n $MON_NS eventrouter
+kubectl delete --ignore-not-found servicemonitor -n $MON_NS elasticsearch
```
gsmith-sas, line 58: Service Monitors are now deployed (and removed) from the logging deployment/removal scripts.  Do we need these here too?

monitoring/openshift/grafana-proxy-secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    k8s-app: grafana
+  name: grafana-proxy
+type: Opaque
+data:
+  session_secret: djRtLW9wZW5zaGlmdC1tb25pdG9yaW5nLXNlY3JldAo=
```
gsmith-sas, line 9: Is this the "real" session secret?  Just a place-holder?

monitoring/openshift/grafana-trusted-ca-bundle.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  creationTimestamp: "2021-05-08T01:44:45Z"
+  labels:
```
gsmith-sas, line 5: Do we need this particular creationTimestamp value (or some hard-coded value)?

gsmith-sas (commented):
A couple of minor comments/questions.

bin/openshift-include.sh
```diff
@@ -0,0 +1,40 @@
...
+# Assumes bin/common.sh has been sourced
+
+# Detect if it is an OpenShift cluster
+OPENSHIFT_CLUSTER=${OPENSHIFT_CLUSTER:-false}
+if kubectl get ns openshift 2>/dev/null 1>&2; then
+  log_debug "OpenShift detected"
+  OPENSHIFT_CLUSTER=true 
+else
+  log_debug "OpenShift not detected. Skipping 'oc' checks."
+fi
```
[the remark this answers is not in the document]
BryanEllington: Not sure what the problem is. The oc check only occurs if OPENSHIFT_CLUSTER==true and OPENSHIFT_OC_CHECK==true. Unset or 'false' result in the same behavior.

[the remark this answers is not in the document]
gsmith-sas: It's a nit but: if someone explicitly set OPENSHIFT_CLUSTER to true but was NOT running on OpenShift, the debug message would indicate that the oc check would be skipped...except that wouldn't happen.  You aren't setting OPENSHIFT_CLUSTER to "false" if you determine you are not running on OpenShift, so nothing is skipped.  The message indicates some change of execution logic will occur due to the failed kubectl call but nothing in the code changes the execution logic.  If you set OPENSHIFT_OC_CHECK to false in the else block, you would be explicitly skipping the oc check.  So, as written, the first part of the debug message is correct, the 2nd is misleading or inaccurate.  But, as I said, this is all a bit of nit.

gsmith-sas (approved):
LGTM

davidbass13 (approved):

PR #123, opened 2021-06-22
Support TLS for logging on OpenShift

gsmith-sas:


logging/bin/deploy_logging_open_openshift.sh
```diff
@@ -38,90 +38,87 @@ log_notice "Deploying logging components to the [$LOG_NS] namespace [$(date)]"
...
-logging/bin/deploy_openshift_routes.sh
+log_info "STEP 4a: Create OpenShift Route(s) & Application URL(s)"
 
-##################################
-# Kibana Content                 #
-##################################
+OPENSHIFT_ROUTES_ENABLE=${OPENSHIFT_ROUTES_ENABLE:-true}
 
-# NOTE: For ODFE, Kibana is deployed as part of ES Helm chart
+if [ "$OPENSHIFT_ROUTES_ENABLE" == "true" ]; then
```
BryanEllington, line 85: If we're going to use a generic env name like this, then it should also be supported on the monitoring side.

logging/bin/deploy_logging_open_openshift.sh
```diff
@@ -38,90 +38,87 @@ log_notice "Deploying logging components to the [$LOG_NS] namespace [$(date)]"
...
 
-log_info "STEP 4b: Configuring Kibana"
+   servicelist="KIBANA"
+   logging/bin/create_openshift_route.sh KIBANA
 
-export KB_KNOWN_NODEPORT_ENABLE=false
-logging/bin/deploy_kibana_content.sh
+   ES_ENDPOINT_ENABLE=${ES_ENDPOINT_ENABLE:-false}
+   if [ "$ES_ENDPOINT_ENABLE" == "true" ]; then
```
BryanEllington, line 91: `ES_ENDPOINT_ENABLE` should probably be `OPENSHIFT_ES_ROUTE_ENABLE` or something similar.

logging/bin/deploy_logging_open_openshift.sh
gsmith-sas: Sounds reasonable to me.  Make it so. ;-)

logging/bin/deploy_logging_open_openshift.sh
gsmith-sas: I started with that but I'd like to support the env var in non-OpenShift environments at some point.  In the sample user.env files, the comments indicate that it is OpenShift specific.

BryanEllington: Well, "endpoint" is the wrong term to use here regardless. Kubernetes' defines an 'endpoint' as one target that a service resolves to. They've since moved to EndpointSlices, but still, let's avoid the use of 'endpoint'.

IMO, if we have `OPENSHIFT_ROUTES_ENABLE`, we should have `OPENSHIFT_ES_ROUTE_ENABLE`.

I think OpenShift is just different. We can have INGRESS_ENABLE in the future for non-OpenShift environments.

gsmith-sas: Check latest push...changed ES_ENDPOINT_ENABLE to OPENSHIFT_ES_ROUTE_ENABLE in code and two user.env files.

BryanEllington (approved):
Thumbs up

PR #124, opened 2021-06-25
Obtain OpenShift server version, enforce client and server version ch…

BryanEllington:
…ecks

### DIFF
--- a/bin/openshift-include.sh
+++ b/bin/openshift-include.sh
@@ -52,19 +52,19 @@ function ocVersionCheck {
       log_debug "OpenShift version check OK"
     elif [ "$OSHIFT_MAJOR_VERSION" -lt 4 ]; then
       log_error "Unsupported OpenShift version: $OSHIFT_FULL_VERSION"
-      log_error "Supported versions include: 4.6.x"
+      log_error "Version 4.6+ is required"
       exit 1
     elif [ "$OSHIFT_MAJOR_VERSION" -eq 4 ] && [ "$OSHIFT_MINOR_VERSION" -lt 6 ]; then
       log_error "Unsupported OpenShift version: $OSHIFT_FULL_VERSION"
-      log_error "Supported versions include: 4.6.x"
+      log_error "Version 4.6+ is required"
       exit 1
     elif [ "$OSHIFT_MAJOR_VERSION" -eq 4 ] && [ "$OSHIFT_MINOR_VERSION" -gt 6 ]; then
       # 4.7+ (not 5+) should still work, so just issue a warning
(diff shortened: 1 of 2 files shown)


bin/openshift-include.sh
```diff
@@ -4,6 +4,73 @@
...
+  if [ "$OPENSHIFT_VERSION_CHECK" == "true" ]; then
+    if [ "$OC_MAJOR_VERSION" -lt 4 ]; then
+      log_error "Unsupported 'oc' version: $OC_FULL_VERSION. Version 4+ is required."
+      exit 1
+    fi
+    if [ "$OSHIFT_MAJOR_VERSION" -eq 4 ] && [ "$OSHIFT_MINOR_VERSION" -eq 6 ]; then
+      log_debug "OpenShift version check OK"
+    elif [ "$OSHIFT_MAJOR_VERSION" -lt 4 ]; then
+      log_error "Unsupported OpenShift version: $OSHIFT_FULL_VERSION"
+      log_error "Supported versions include: 4.6.x"
```
gsmith-sas, line 55: Why the vague wording?  You're throwing an error if it isn't at least 4.6...seems like the supported version is 4.6+

bin/openshift-include.sh
```diff
@@ -4,6 +4,73 @@
...
+            echo "Unable to parse client version: [$ver]"
+          fi
+      elif [ "$verType" == "Server" ]; then
+          ver="${verArr[2]}"
+          if [[ $ver =~ (([0-9]+)\.([0-9]+)\.([0-9]+)) ]]; then
+            OSHIFT_FULL_VERSION=${BASH_REMATCH[1]}
+            OSHIFT_MAJOR_VERSION=${BASH_REMATCH[2]}
+            OSHIFT_MINOR_VERSION=${BASH_REMATCH[3]}
+            OSHIFT_PATCH_VERSION=${BASH_REMATCH[4]}
+          else
```
gsmith-sas, line 37: Looks like an opportunity for consolidating code or creating a semver parse function....but we may never need that again. Probably not worth the effort.

gsmith-sas (approved):
I made some comments but don't think any _require_ action.  I think the messages about supported OC version should be less wishy-washy, but, again, not critical.  Fix or push as you see fit.

bin/openshift-include.sh
BryanEllington: I think a refactor is in the cards for the future

bin/openshift-include.sh
BryanEllington: See latest changes

gsmith-sas (approved):
LGTM.

PR #128, opened 2021-06-29
First pass of path-based route support for OpenShift

BryanEllington:


samples/generic-base/monitoring/user.env
```diff
@@ -56,6 +56,13 @@
 # LOGGING_DASH=true
 # USER_DASH=true
 
+# OpenShift-specific options
+# Used when OPENSHIFT_PATH_ROUTES=true
+# Fully-qualified route hostname used for Grafana
+# OPENSHIFT_ROUTE_HOST_GRAFANA
```
gsmith-sas, line 62: Should we include a sample value here?  Or, iff that's not possible, an obvious non-sense value?  As it is now, there is just a variable name with no assignment operator or value...not at all clear that this is a value that should/can be changed.

samples/generic-base/user.env
```diff
@@ -20,10 +20,16 @@
...
 # Name of NGINX ingress controller (if applicable)
 # NGINX_SVCNAME=ingress-nginx-controller
 
+# OpenShift-specific options
+# Enable path-based routes (URLs will end in /grafana, /kibana, etc.)
+# OPENSHIFT_PATH_ROUTES=false
+# Override the base route hostname used for monitoring and logging
+# Note that the actual hostname used will have the name of the service
+# and namespace prepended (e.g. v4m-grafana-$MON_NS.$OPENSHIFT_ROUTE_DOMAIN)
+# OPENSHIFT_ROUTE_DOMAIN
```
gsmith-sas, line 35: See comment above about providing a sample value or an obvious non-sense value.  As written, it is not clear that this is a value the user can/should change.

gsmith-sas (commented):
Had some questions/suggestions...

monitoring/bin/deploy_monitoring_openshift.sh
```diff
@@ -168,9 +187,25 @@ for f in monitoring/rules/viya/rules-*.yaml; do
...
 if ! kubectl get route -n $MON_NS v4m-grafana 1>/dev/null 2>&1; then
   log_info "Exposing Grafana service as a route..."
-  oc create route reencrypt -n $MON_NS --service=v4m-grafana
+  if [ "$OPENSHIFT_PATH_ROUTES" == "true" ]; then
+    oc create route reencrypt \
+      -n $MON_NS \
+      --service v4m-grafana \
+      --hostname "$routeHost" \
+      --path $OPENSHIFT_ROUTE_PATH_GRAFANA
+  else
```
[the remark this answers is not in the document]
BryanEllington: Maybe in a separate, future PR. Right now, it's written as a standalone script and only supports logging components. I don't think refactoring now is the right thing to do.

monitoring/bin/deploy_monitoring_openshift.sh
```diff
@@ -182,7 +217,13 @@ if [ ! "$OPENSHIFT_AUTH_ENABLE" == "true" ]; then
...
 fi
-log_notice "Grafana URL: $scheme://$(kubectl get route -n $MON_NS | grep v4m-grafana | awk '{printf $2}')"
+bin/show_app_url.sh GRAFANA
+# log_notice "Grafana URL: $scheme://$(kubectl get route -n $MON_NS v4m-grafana -o jsonpath='{.spec.host}{.spec.path}')"
 
 log_message ""
 log_notice "Successfully deployed SAS Viya monitoring for OpenShift"
+
+function url {
+  scheme=$(figure out tls present???)
```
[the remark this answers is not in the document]
BryanEllington: Oops. Leftover junk that should have been deleted. Will remove.

monitoring/openshift/grafana-proxy-patch-path.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - configMap:
+          defaultMode: 420
+          items:
+          # - key: ca.crt
+          - key: ca-bundle.crt
+            path: tls-ca-bundle.pem
+          # name: kube-root-ca.crt
+          name: grafana-trusted-ca-bundle
+          optional: true
+        name: grafana-trusted-ca-bundle
```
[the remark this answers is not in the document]
BryanEllington: The patch adds a sidecar container and some volumes. Those structures are stable in the helm chart, so I'm not too worried.

samples/generic-base/monitoring/user.env
BryanEllington: Updated

samples/generic-base/user.env
BryanEllington: Updated

gsmith-sas (approved):
LGTM

PR #132, opened 2021-07-06
Add path-based info for customizing OpenShift routes

BryanEllington:


logging/OpenShift.md
```diff
@@ -111,6 +111,21 @@ oc -n $LOG_NS delete route v4m-es-client-service
 ***Note:*** Do not use the `es_nodeport_enable.sh` and `es_nodeport_enable.sh` scripts 
 in an OpenShift environment.  
 
+#### Customizing Routes
+
+Routes may be configured to be host-based (default) or path-based. When
+using host-based routes, access to logging applications will be by hostname
+only - for example: `https://v4m-kibana-monitoring.apps.my-openshift-cluster.com`.
```
gsmith-sas, line 118: I'm not sure "_...by hostname only_" is too clear.  Maybe, something like _"When using host-based routes, the name of the application is part of the hostname in the URL (for example:...).  When using path-based routes, the application name only appears as part of the path at the end of the URL"_

logging/OpenShift.md
```diff
@@ -111,6 +111,21 @@ oc -n $LOG_NS delete route v4m-es-client-service
...
+Path-based routes will result in a url ending in `/appName` - for example:
+`https://v4m-logging.apps.my-openshift-cluster.com/kibana`.
+
+Use `OPENSHIFT_PATH_ROUTES=true` in `$USER_DIR/user.env` to use path-based
+routes.
+
+The hostname can be configured using `OPENSHIFT_ROUTE_HOST_KIBANA` and/or
+`OPENSHIFT_ROUTE_HOST_ELASTICSEARCH` (if the Elasticsearch route is enabled)
+in `$USER_DIR/user.env`.
```
gsmith-sas, line 127: Should we clarify what the default hostname will be if users do NOT specify values for these env vars?

logging/OpenShift.md
```diff
@@ -111,6 +111,21 @@ oc -n $LOG_NS delete route v4m-es-client-service
...
+
+Routes may be configured to be host-based (default) or path-based. When
+using host-based routes, access to logging applications will be by hostname
+only - for example: `https://v4m-kibana-monitoring.apps.my-openshift-cluster.com`.
+Path-based routes will result in a url ending in `/appName` - for example:
+`https://v4m-logging.apps.my-openshift-cluster.com/kibana`.
+
+Use `OPENSHIFT_PATH_ROUTES=true` in `$USER_DIR/user.env` to use path-based
+routes.
+
```
gsmith-sas, line 124: They could define this env var in the $USER_DIR/logging/user.env file instead of the top-level user.env file too; correct?  If so, that would limit its impact to the logging applications only.  Should we mention that option?  I'm reluctant to document every possible combination, but given that this document is in the logging sub-directory, maybe we need to touch on it.

logging/OpenShift.md
BryanEllington: I give an example

logging/OpenShift.md
BryanEllington: I'll leave the wording to @davidbass13

logging/OpenShift.md
BryanEllington: It's our standard customization...seems unnecessary to re-hash how `user.env` files works in a `$USER_DIR`. I'll defer to @davidbass13 .

logging/OpenShift.md
davidbass13: I agree with the wording suggested by @gsmith-sas, although I would add an example for the path-based URL as well.

logging/OpenShift.md
davidbass13: I think it wouldn't hurt to mention it. I'd word it as 'Specify 'OPENSHIFT_PATH_ROUTES=true' in the '$USER_DIR/user.env' file (for both logging and monitoring) or the '$USER_DIR/logging/user.env' file (for only logging components) to use path-based routes.'

davidbass13 (commented):
Looks fine, with the suggestions made by @gsmith-sas

BryanEllington (commented):
Docs updated. Please look at the revised versions

davidbass13:
Changes look good

PR #139, opened 2021-07-12
Tweaks to allow (re-)running log deployment scripts on OpenShift

gsmith-sas:


logging/bin/create_openshift_route.sh
```diff
@@ -63,10 +63,8 @@ case "$app" in
 esac
 
 if oc -n $namespace get route $route_name 2>/dev/null 1>&2; then
-   log_error "The requested route [$route_name] already exists in the namespace [$namespace]."
-   log_error "The details of existing route are shown below:"
-   oc -n $namespace get route $route_name
-   exit 1
+   log_info "The requested route [$route_name] already exists in the namespace [$namespace]."
```
BryanEllington, line 66: I think adding something like "Skipping route creation. The requested route ..." would help clarify the purpose of this message.

gsmith-sas: Good point.  Updated wording in both scripts.

PR #146, opened 2021-07-21
[FIX] Disable/Re-enable ES NodePort (as-needed) during redeployment

gsmith-sas:


BryanEllington (commented):
I'm a little confused about es_nodeport_enable_open.sh - it has a check for whether ES_PORT is != 0, but it keeps going and enables the port regardless. Shouldn't there be an exit there?

BryanEllington (approved):

PR #147, opened 2021-07-30
Add helm-based version info

BryanEllington:


bin/version-include.sh
```diff
@@ -0,0 +1,104 @@
...
+    cat "$USER_DIR/logging/user.env" | sed 's/^/      /' >> "$v4mValuesYAML"
+  fi
+}
+
+function deployV4MInfo() {
+  NS=$1
+  if [ -z "$NS" ]; then
+    log_error "No namespace specified for deploying version info"
+    return 1
+  fi
```
gsmith-sas, line 49: I don't see any logic in the calling programs to deal with a non-zero return code.  Do we want them to error out at that point?  Generate a message?  Do nothing?

bin/version-include.sh
```diff
@@ -0,0 +1,104 @@
...
+        V4M_CURRENT_VERSION_FULL=${BASH_REMATCH[1]}
+        V4M_CURRENT_VERSION_MAJOR=${BASH_REMATCH[2]}
+        V4M_CURRENT_VERSION_MINOR=${BASH_REMATCH[3]}        
+        V4M_CURRENT_VERSION_PATCH=${BASH_REMATCH[4]}
+      elif [[ "$line" =~ $sre ]]; then
+        V4M_CURRENT_STATUS=${BASH_REMATCH[1]}
+      fi
+    done
+
+    log_debug "V4M_CURRENT_VERSION_FULL=$V4M_CURRENT_VERSION_FULL"
```
gsmith-sas, line 92: If there is no previous release found, do we still want all of these debug messages with blank values?

v4m-chart/values.yaml
```diff
@@ -0,0 +1 @@
+# Default values for v4m
```
gsmith-sas, line 1: This file is never actually used, is it?  If a user adds values here, are we using them in some way?  If has to be here or if we want it here b/c that is "best practices" for Helm charts...we should probably include comments indicating that there are no default values supported and the file isn't actually used.

gsmith-sas (commented):
I think this looks pretty good...was a little confused about the values.yaml file included and whether it is actually used...and how we should handle non-zero return values from the functions.

v4m-chart/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: v4m
+description: Viya 4 Monitoring for Kubernetes (https://github.com/sassoftware/viya4-monitoring-kubernetes)
+type: application
+version: 1.0.10-SNAPSHOT
+appVersion: "1.0.10-SNAPSHOT"
```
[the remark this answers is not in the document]
BryanEllington: Yes

v4m-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "v4m.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "v4m.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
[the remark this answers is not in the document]
BryanEllington: I think we should make a full pass at this sometime. This particular bit only affects configMap deployed by the skeleton chart.

bin/version-include.sh
```diff
@@ -0,0 +1,104 @@
...
+    --values $valuesYAML \
+    v4m ./v4m-chart
+}
+
+function removeV4MInfo() {
+  NS=$1
+  if [ -z "$NS" ]; then
+    log_error "No namespace specified for removing version info"
+    return 1
+  fi
```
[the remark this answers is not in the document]
BryanEllington: Fixed

bin/version-include.sh
BryanEllington: Fixed

bin/version-include.sh
BryanEllington: The if block above handles the case where there is no existing release and this info will not get output.

v4m-chart/values.yaml
BryanEllington: It's a convention that there is a default values.yaml file. We may use it in the future, but yes, for now, it's empty. I added a comment.

v4m-chart/Chart.yaml
```diff
@@ -0,0 +1,7 @@
...
+# foo
```
gsmith-sas, line 7: What's with the # foo line?

gsmith-sas (approved):
Am confused by the # foo line added in one file but otherwise looks good to me.

v4m-chart/Chart.yaml
BryanEllington: oops. was testing the new 'git status' functionality to detect changes to the local repo

PR #149, opened 2021-08-05
Revised TLS docs for logging to make differences clearer

davidbass13:


logging/Notes_on_using_TLS.md
```diff
@@ -2,7 +2,21 @@
...
+- **TLS for communications into the cluster**, which is between a user's 
+browser and Kibana. TLS is optional on these connections, although enabling
+TLS is a best practice.
+
+The information in this document explains only how to configure TLS 
+for in-cluster communications, either automatically (using cert-manager) 
+or manually.
+
+For information about configuring TLS for communications into the cluster, see 
+the [TLS logging sample](../samples/tls/loggingREADME.md).
```
gsmith-sas, line 20: I wonder whether it might be more clear to move these two paragraphs to immediately follow the category of TLS enablement they address rather than introduce the two categories and then follow it with two more paragraphs that each address one of the categories.
So, organize things like:

- TLS for in-cluster communications
- This document explains how to...
- TLS for communications into the cluster
- For information about configuring TLS for communication into the cluster, see...

As it is currently organized, the reader has to think about A then B and then back to A and then back to B.

logging/Notes_on_using_TLS.md
```diff
@@ -2,7 +2,21 @@
...
+- **TLS for in-cluster communications**, which is between logging components within
+the cluster (between the Elasticsearch nodes and between Elasticsearch and other logging components). TLS must be enabled on these connections.
+
+- **TLS for communications into the cluster**, which is between a user's 
+browser and Kibana. TLS is optional on these connections, although enabling
+TLS is a best practice.
+
+The information in this document explains only how to configure TLS 
+for in-cluster communications, either automatically (using cert-manager) 
+or manually.
```
gsmith-sas, line 16: The automatic (cert-manager)/manual steps aren't really configuring TLS, they're generating/ensuring the required TLS certificates  are in-place to enable TLS to work.  Might be a nit but cert-manager really doesn't configure anything...the deployment scripts handle the actual configuration work, cert-manager is called upon to generate the certificates.

logging/Notes_on_using_TLS.md
```diff
@@ -2,7 +2,21 @@
...
 
-TLS is required for communication between the Elasticsearch nodes and for accessing Elasticsearch from other components. It is also a best practice to use TLS (by using HTTPS) when accessing Kibana from a browser.
+TLS enablement for SAS Viya logging is divided into two parts:
+
+- **TLS for in-cluster communications**, which is between logging components within
+the cluster (between the Elasticsearch nodes and between Elasticsearch and other logging components). TLS must be enabled on these connections.
+
+- **TLS for communications into the cluster**, which is between a user's 
+browser and Kibana. TLS is optional on these connections, although enabling
+TLS is a best practice.
```
gsmith-sas, line 12: Technically, enabling TLS here isn't limited to the browser->Kibana use-case.  This also covers the case of requiring TLS between ingress and Kibana (if ingress is configured).  There is also a use-case for connecting to the Elasticsearch API which requires its own TLS cert.

samples/tls/logging/README.md
```diff
@@ -2,7 +2,24 @@
...
+
+This sample demonstrates only how to enable TLS for communications into the 
+cluster by deploying logging with TLS enabled for connections between the 
+user and Kibana. The `TLS_ENABLE` environment variable controls whether 
+connections to Kibana use TLS.
+
+In-cluster communications must always take place through TLS-enabled connections. TLS for in-cluster communications is configured using a separate process, 
+either automatically (using cert-manager) or manually. See [Notes_on_using_TLS](../../../logging/Notes_on_using_TLS.md) for information about configuring TLS for 
+in-cluster communications.
+
```
gsmith-sas, line 22: See note above about reorganizing the paragraphs...think it makes sense to do that here as well.

samples/tls/logging/README.md
```diff
@@ -22,39 +39,7 @@ my_repository_path/logging/bin/deploy_logging_open.sh
...
-kubectl create secret tls $SECRET_NAME -n $NAMESPACE --key $CERT_KEY --cert $CERT_FILE
-```
-
-Use `kibana-ingress-tls-secret` and `elasticsearch-ingress-tls-secret` as values for `$SECRET_NAME`. Use `logging` for the value of `$NAMESPACE`.
-
-The process of generating the certificates for these secrets is out of scope for this example.
-
-* If you are using an ingress controller other than NGINX, modify the annotation 
-`nginx.ingress.kubernetes.io/backend-protocol: HTTPS` as needed in the `user-values-elasticsearch-open.yaml` file. Refer to the documentation for your ingress controller. 
-
```
gsmith-sas, line 48: I think the Ingress section needs to be kept.  It is NOT covered in the NOTES_ON_TLS.md doc.

gsmith-sas (commented):
I think the new intro discussion of the two forms of TLS is good...but could be tweaked.  I also think the discussion of the ingress TLS certs was accidentally cut from the samples file.  I'm going to ask Bryan to take a look at the this as well.  I think (fear?) yet another phone call may be necessary to make sure we cover things appropriately.

BryanEllington (commented):
I think we should use the term "ingress" more than generic phrases like "communications into the cluster" and "incoming traffic".

samples/tls/logging/README.md
```diff
@@ -7,20 +7,19 @@ TLS enablement for SAS Viya logging is divided into two parts:
...
-- **TLS for communications into the cluster**, which is between a user's 
-browser and Kibana. TLS is optional on these connections, although enabling
-TLS is a best practice.
+In-cluster communications must always take place through TLS-enabled connections. TLS for in-cluster communications is configured using a separate process, 
+either automatically (using cert-manager to generate certificates) 
+or by manually generating certificates. See [Notes_on_using_TLS](../../../logging/Notes_on_using_TLS.md) for information about configuring TLS for 
+in-cluster communications.
+
+- **TLS for communications into the cluster**, which is between Ingress or a 
+user's browser (if NodePorts are used) and Kibana. TLS is optional on these connections, although enabling TLS is a best practice.
```
gsmith-sas, line 17: Revise the summary for the corresponding section (one TLS for communications into the cluster) in NOTE ON USING TLS to match this one.

samples/tls/logging/README.md
```diff
@@ -37,9 +36,51 @@ After you finish modifying the customization files, deploy logging using the sta
...
+
+2. After you have obtained the certificates for the secrets, use this command to generate the secrets:
+
+```bash
+kubectl create secret tls $SECRET_NAME -n $NAMESPACE --key $CERT_KEY --cert $CERT_FILE
+```
+
+Use `elasticsearch-ingress-tls-secret` and `kibana-ingress-tls-secret` as values for `$SECRET_NAME`. Use the name of the namespace into which the logging components 
+were deployed (such as `logging`) for the value of `$NAMESPACE`.
+
```
gsmith-sas, line 78: Steps 1 and 2 are doing the same thing (i.e. creating the secret from the certs).  I notice the syntax is different...go with what is there in Step 2 since that is pulled (I''m pretty sure) from the original version of this particular doc.  
Since the syntax used in the previous section is different, it might confuse the users.  I suggest rather than showing it here, you just point them to the NOTES ON TLS doc for the details of creating the kibana-tls-secret secret if they want to do that manually.

samples/tls/logging/README.md
```diff
@@ -37,9 +36,51 @@ After you finish modifying the customization files, deploy logging using the sta
...
+3. In your local copy of the file 
+`$USER_DIR\samples\tls\logging\user-values-elasticsearch-open.yaml`, replace the 
+values `kibana.logging.host.cluster.example.com` and `elasticsearch.logging.host.cluster.example.com` with valid locations for your environment.
+
```
gsmith-sas, line 82: They need to modify those files to provide valid hostnames and paths for their environments (i.e. portions of the URL used to access the applications) rather than "locations" which might be confused with file/directory paths.

gsmith-sas (commented):
I think the revision look good.  I, of course, had a couple of tweaks (see comments).  Also, I don't think we list the kibana-tls-secret in the list of secrets in the Notes_on_using_TLS.md document.  Those TLS certs aren't really used unless TLS_ENABLE is set to true (which is what the mark-down file in the samples/tls sub-dir covers) but we should probably include it since the secret MUST be there for the deployment process to work...even if the certs aren't actually used.

samples/tls/logging/README.md
gsmith-sas: Still not comfortable with the use of "locations" here.   Maybe something like: _"In your local copy of the file...update the  the values `kibana.logging.host.cluster.example.com` and `elasticsearch.logging.host.cluster.example.com` to reflect the namespace in which you've deployed the log monitoring components and the correct ingress host information."_

gsmith-sas (changes requested):
I like the revisions, I think they do a good job of conveying the right information and pointing readers to the appropriate places.  I'm still not comfortable with the use of the word "locations" in Step 3 of the samples readme...I've suggested alternative wording to help clarify(?) my concern.

davidbass13:
Thanks for the suggested wording, @gsmith-sas. Changes made

gsmith-sas (approved):
LGTM

PR #161, opened 2021-09-02
Initial deployment and removal of tenant monitoring

BryanEllington:


monitoring/bin/deploy_monitoring_tenant.sh
```diff
@@ -0,0 +1,170 @@
...
+  $extraArgs \
+  grafana-$VIYA_TENANT \
+  grafana/grafana
+
+# Deploy ServiceMonitors
+kubectl apply -n $VIYA_NS -f $tenantDir/serviceMonitor-sas-cas-tenant.yaml
+kubectl apply -n $VIYA_NS -f $tenantDir/serviceMonitor-sas-pushgateway-tenant.yaml
+
+function deploy_tenant_dashboards {
+   dir=$1
```
gsmith-sas, line 136: Possible future refactoring opportunity: call the (existing) monitoring/bin/deploy_dashboards.sh script after supporting any tenant-specific needs.

monitoring/bin/deploy_monitoring_tenant.sh
```diff
@@ -0,0 +1,170 @@
...
+  # Find the grafana pod
+  grafanaPod="$(kubectl get po -n $VIYA_NS -l app.kubernetes.io/name=grafana --template='{{range .items}}{{.metadata.name}}{{end}}')"
+
+  log_notice "====================================================================="
+  log_notice "Generated Grafana admin password is: $grafanaPwd"
+  log_notice "Change the password at any time by running (replace password):       "
+  log_notice "kubectl exec -n $VIYA_NS $grafanaPod -c grafana -- bin/grafana-cli admin reset-admin-password myNewPassword"
+  log_notice "====================================================================="
+  log_message ""
+fi
```
gsmith-sas, line 168: Doesn't look like we show the URL for the tenant-specific Grafana like we do when running deploy_monitoring_cluster.sh; should we?

gsmith-sas (approved):
I was surprised the deployment script does not display the Grafana URL.  That seems like an important bit of info and it is something displayed by the cluster deployment script.  Might be good to add it (either now or later).

Nit: In the cas-dashboard.json I see the description is set to"SAS Viya CAS dashboard prototype".  Do we want/need the word "prototype" there?

PR #174, opened 2021-09-14
Cert_doc

davidbass13:
Addressing reported documentation error. NOTES_ON_USING_TLS needs to indicate that different certificates are needed for each of the Kubernetes secrets that are manually generated.


gsmith-sas (commented):
This looks good to me.  My only comment is that I wonder if we need to add a pointer to the TLS sample in the ingress sample readme.  This would ensure that people doing ingress are aware of the potentially related TLS issues/requirements.

gsmith-sas (approved):
Looks good to me.  Thanks!

PR #178, opened 2021-09-21
Monitoring TLS support for tenant onboarding

BryanEllington:


monitoring/multitenant/mt-prometheus.yaml
```diff
@@ -1,33 +1,38 @@
...
 apiVersion: v1
-kind: ServiceAccount
+kind: Service
 metadata:
-  name: prometheus-__TENANT__
   labels:
+    app: prometheus
+    self-monitor: "true"
     sas.com/tenant: __TENANT__
+  name: prometheus-__TENANT__
```
gsmith-sas, line 8: Hmmm, not sure we want to use this label here.  SEE TOP-LEVEL REVIEW COMMENT.

monitoring/multitenant/mt-prometheus.yaml
```diff
@@ -1,33 +1,38 @@
...
-kind: RoleBinding
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
 metadata:
-  name: prometheus-rolebinding-__TENANT__
-subjects:
-- kind: ServiceAccount
+  labels:
+    app: prometheus
+    sas.com/tenant: __TENANT__
```
gsmith-sas, line 26: See comment above re: use of the sas.com/tenant label.

gsmith-sas (commented):
My only concern is the use of the sas.com/tenant Kubernetes label.  The sas.com/tenant label was created to identify tenant-specific pods so metrics and logs from them could be made available to the corresponding tenant admins. While I can see that you are applying the label to non-pod resources here, I still question whether the label is appropriate here. Are metrics, log messages or anything else related to these resource intended for the corresponding tenant admin? If not, I don't think the label should be used here. We want to avoid over-using the label or using it for purposes other than its original intent (even if the label "makes sense" since the resource is related to a tenant).  The label is not being used as a "general-purpose" tenant-identifier on the SAS Viya side, therefore, I don't think we should use it that way either.

gsmith-sas (approved):
LGTM

PR #187, opened 2021-10-04
Revised ES pipeline to only direct 'primary' container logs to tenant…

gsmith-sas:
… indexes


BryanEllington (approved):
Not an expert in Elasticsearch pipelines, but it looks correct

PR #190, opened 2021-10-06
Require kubectl 1.18+ as pre-requisite.

cumcke:


monitoring/bin/deploy_dashboards.sh
```diff
@@ -23,18 +23,6 @@ TEST_DASH="${TEST_DASH:-false}"
...
-  dryRun="--dry-run"
-elif [[ $KUBE_CLIENT_VER =~ v1.1[8-9] ]]; then
-  dryRun="--dry-run=client"
-elif [[ $KUBE_CLIENT_VER =~ v1.[2-9] ]]; then
-  dryRun="--dry-run=client"
-else 
-  log_error "Unsupported kubectl version: [$KUBE_CLIENT_VER]"
-  exit 1
-fi
-
```
BryanEllington, line 37: It's fine to remove this block, but you need to replace the uses of $dryRun with `--dry-run=client`

BryanEllington (changes requested):
It looks like the block under "# The kubectl --dry-run command changed as of v1.18" is back. That block that sets the dryRun variable should be removed now that the value is hard-coded.

Also, you need to replace the 2nd use of `$dryRun` on line 71 with `--dry-run=client`

cumcke (commented):
Suggested changes have been made.

BryanEllington (changes requested):
Still missing a change on line 59 of deploy_dashboards.sh - Replace `$dryRun` with `--dry-run=client`

gsmith-sas (commented):
Looks good to me but I will defer to Bryan for approval.

PR #212, opened 2021-10-27
Move to Open Distro for Elasticsearch 1.13.2

gsmith-sas:


logging/bin/deploy_elasticsearch_open.sh
```diff
@@ -185,14 +214,12 @@ if [ ! -f "$TMP_DIR/$odfe_tgz_file" ]; then
...
 
-   # checkout specific commit
    cd opendistro-build
-   git checkout 2c139044ee31a490b58ac6a306a5a5ef5ef21383
 
    # build package
    log_info "Packaging Helm Chart for Elasticsearch"
 
-   # cd opendistro-build/helm/opendistro-es/
+   # cd to the appropriate sub-directory
```
BryanEllington, line 222: Comment isn't very helpful..maybe just remove it.

logging/bin/getlogs.sh
```diff
@@ -33,7 +33,7 @@ default_maxrows=500
 
 # NOTE: the 'trim(foo.bar.keyword) as bar' syntax is needed to ensure nested fields appear
 #       as properly named columns rather than as nested fields "(e.g. foo={'bar'='baz'})"
-default_output_vars="@timestamp, level, logsource,kube.namespace, kube.pod, kube.container, message"
+#default_output_vars="@timestamp, level, logsource,kube.namespace, kube.pod, kube.container, message"
```
BryanEllington, line 36: delete comment?

logging/es/odfe/es_helm_values_open.yaml
```diff
@@ -231,9 +231,9 @@ kibana:
     elasticsearch.password: ${ELASTICSEARCH_PASSWORD}
 
     opendistro_security.multitenancy.enabled: true
+    # opendistro_security.multitenancy.tenants.preferred: ["Global","Private"]
```
BryanEllington, line 234: Delete comment

logging/bin/deploy_elasticsearch_open.sh
```diff
@@ -117,6 +118,34 @@ if [ "$(helm -n $LOG_NS list --filter 'odfe' -q)" == "odfe" ]; then
...
    existingODFE="true"
 
+   #Migrate Kibana content if upgrading from ODFE 1.7.0 to 1.13.2
+   if [ "$(helm -n logging list -o yaml --filter odfe |grep app_version)" == "- app_version: 1.8.0" ]; then
+
+      # Prior to 1.1.0 we used ODFE 1.7.0
+      log_info "Migrating from Open Distro for Elasticsearch 1.7.0"
+
+      #export exisiting content from global tenant
+      #KB_GLOBAL_EXPORT_FILE="$TMP_DIR/kibana_global_content.ndjson"
```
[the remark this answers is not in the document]
gsmith-sas: The file should (if things go right) contain the exported Kibana objects.  But, if things fail, it contains potentially helpful error information.  I suppose a failed migration means the file will be created (signaling a migration)  even if the content migration step fails.  That would signal the migration import processing in deploy_kibana_content.sh to execute...and fail.  Hmmm.  But I think that's what we would want: the existence of the file indicates a migration is taking place...which we want to react to even if we can't migrate the content since there is other processing that needs to happen during a migration (other than migrating content).  At this point, I don't think a failure to export existing content should stop the migration process.  We don't really support an alternative.  I suppose we could abort and force the user to roll back to a previous release but I'm not sure that is needed at this point (when most/all of the existing Kibana content is probably our default stuff anyway).

logging/bin/deploy_elasticsearch_open.sh
```diff
@@ -175,7 +204,7 @@ create_secret_from_file securityconfig/tenants.yml         security-tenants
 # Open Distro for Elasticsearch
 log_info "Deploying Open Distro for Elasticsearch"
 
-odfe_tgz_file=opendistro-es-1.8.0.tgz
+odfe_tgz_file=opendistro-es-1.13.2.tgz
```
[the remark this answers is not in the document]
gsmith-sas: We have a story in the backlog to recode this so it does not include a hard-coded version number.  And I have also suggested to the ODFE developers that the file name shouldn't really include the release/version number.  I've also suggested that they use a chart repo so we don't have to build the chart before using it.  But, as you know, ODFE 1.13.2 is the end of the line for ODFE, so they aren't making changes on their side.  Bottom line: I don't think the hard-coded reference is a problem here.  I'm hoping they improve how they handle this sort of thing in OpenSearch.

logging/bin/deploy_elasticsearch_open.sh
gsmith-sas: Will remove it.

logging/bin/es_nodeport_disable_open.sh
```diff
@@ -12,5 +12,11 @@ export SVC=v4m-es-client-service
...
 log_info "Removing NodePort for Elasticsearch instance [$SVC] in [$LOG_NS]"
 
-kubectl -n "$LOG_NS" patch svc "$SVC" --type='json' -p '[{"op":"replace","path":"/spec/type","value":"ClusterIP"},{"op":"remove","path":"/spec/ports/0/nodePort"},{"op":"remove","path":"/spec/ports/1/nodePort"},{"op":"remove","path":"/spec/ports/2/nodePort"}]' 
+if [[ $KUBE_SERVER_VER =~ v1.18 ]]; then
+   kubectl -n "$LOG_NS" patch svc "$SVC" --type='json' -p '[{"op":"replace","path":"/spec/type","value":"ClusterIP"},{"op":"remove","path":"/spec/ports/0/nodePort"},{"op":"remove","path":"/spec/ports/1/nodePort"},{"op":"remove","path":"/spec/ports/2/nodePort"}]' 
+elif [[ $KUBE_SERVER_VER =~ v1.19 ]]; then
+   kubectl -n "$LOG_NS" patch svc "$SVC" --type='json' -p '[{"op":"replace","path":"/spec/type","value":"ClusterIP"},{"op":"remove","path":"/spec/ports/0/nodePort"},{"op":"remove","path":"/spec/ports/1/nodePort"},{"op":"remove","path":"/spec/ports/2/nodePort"},{"op":"remove","path":"/spec/ports/3/nodePort"}]' 
+else
+   kubectl -n "$LOG_NS" patch svc "$SVC" --type='json' -p '[{"op":"replace","path":"/spec/type","value":"ClusterIP"}]'
+fi
```
gsmith-sas, line 22: @BryanEllington  Can you take a look at this please.  K8s 1.19.x adds an additional port to be removed, so we were failing on some clusters. 😞  And K8s 1.20 added the following (which makes things much easier/better):
> Certain fields on Service objects will be automatically cleared when changing the service's type to a mode that does not need those fields. For example, changing from type=LoadBalancer to type=ClusterIP will clear the NodePort assignments, rather than forcing the user to clear them. (#95196, @thockin) [SIG API Machinery, Apps, Network and Testing]
--from [Kubernetes 1.20 release notes](https://v1-20.docs.kubernetes.io/docs/setup/release/notes/)

BryanEllington: It's nice that 1.20+ automatically handles this annoying issue, but we still need to support 1.19. The code looks correct to me.

logging/bin/deploy_elasticsearch_open.sh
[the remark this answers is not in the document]
BryanEllington: I was just more concerned about the export and troubleshooting data getting lost since we delete TMP_DIR by default.

PR #223, opened 2021-11-12
Changing log levels of various script messages.

cumcke:


bin/log-include.sh
```diff
@@ -12,10 +12,23 @@ if [ "$logVerbose" != "true" ]; then
   exec >/dev/null
 fi
 
-function add_notice {
+function add_notice_store {
```
gsmith-sas: Why the new/additional function?  Does the following not work?
`echo "$1$(printf %$(eval 'echo $n')s |tr ' ' ' ')"  >> $TMP_DIR/notices.txt`

logging/bin/deploy_esexporter.sh
```diff
@@ -31,7 +33,7 @@ fi
 # get credentials
 get_credentials_from_secret metricgetter
 rc=$?
-if [ "$rc" != "0" ] ;then log_info "RC=$rc"; exit $rc;fi
+if [ "$rc" != "0" ] ;then log_verbose "RC=$rc"; exit $rc;fi
```
gsmith-sas, line 37: I think can/should be log_debug.  Not sure why it was ever log_info.

logging/bin/deploy_fluentbit_open.sh
```diff
@@ -33,7 +34,7 @@ fi
 # get credentials
 get_credentials_from_secret logcollector
 rc=$?
-if [ "$rc" != "0" ] ;then log_info "RC=$rc"; exit $rc;fi
+if [ "$rc" != "0" ] ;then log_verbose "RC=$rc"; exit $rc;fi
```
gsmith-sas, line 38: see note above...change to log_debug.

logging/bin/deploy_fluentbit_open.sh
```diff
@@ -92,7 +91,7 @@ else
    # use copy in repo
    FB_CONFIGMAP="logging/fb/fluent-bit_config.configmap_open.yaml"
 fi
-log_info "Using FB ConfigMap:" $FB_CONFIGMAP
+# log_info "Using FB ConfigMap:" $FB_CONFIGMAP
```
gsmith-sas, line 95: The message should be written out at some level...probably log_verbose or log_debug.

logging/bin/deploy_logging_open.sh
```diff
@@ -40,28 +40,28 @@ log_notice "Deploying logging components to the [$LOG_NS] namespace [$(date)]"
 
 # Collect Kubernetes events as pseudo-log messages?
 
-log_info "STEP 1: Event Router"
+# log_info "STEP 1: Event Router"
 logging/bin/deploy_eventrouter.sh
```
gsmith-sas, line 44: Is this being kept as a comment for people reviewing the code or are we still undecided if we want to write these messages out?  If it is the former, we can probably remove the function call.  And, the same applies to the rest of the "Step X" messages/comments below.

logging/bin/secrets-include.sh
```diff
@@ -14,7 +14,7 @@ function create_secret_from_file {
 
   if [ -z "$(kubectl -n $LOG_NS get secret $secret_name -o name 2>/dev/null)" ]; then
 
-    log_debug "Will attempt to create secret [$secret_name]"
+    # log_debug "Will attempt to create secret [$secret_name]"
```
gsmith-sas, line 18: I vote for keeping this debug comment in.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -204,14 +205,14 @@ kubectl apply -n $MON_NS -f monitoring/monitors/logging/serviceMonitor-fluent-bi
...
+log_verbose "Adding Prometheus recording rules..."
 for f in monitoring/rules/viya/rules-*.yaml; do
   kubectl apply -n $MON_NS -f $f
 done
 
 kubectl get prometheusrule -n $MON_NS v4m-kubernetes-apps 2>/dev/null
 if [ $? == 0 ]; then
-  log_info "Patching KubeHpaMaxedOut rule..."
+  log_verbose "Patching KubeHpaMaxedOut rule..."
   # Fixes the issue of false positives when max replicas == 1
```
gsmith-sas, line 216: Do we want to keep the "..." at the end of this message?  We have removed them from other messages in this same script.

gsmith-sas (commented):
Looks pretty good...had a few questions/comments.  There were pushes made while I was in the middle of my review...so you may have addressed some of my concerns already.

bin/log-include.sh
cumcke: The old **add_notice_wide** referenced **add_notice** and I was trying to preserve that.
Bryan helped me fix it this morning and should be addressed in a later push.

logging/bin/secrets-include.sh
cumcke: I'll go ahead and keep them in and change the verbiage a bit from "Will attempt to create" to "Creating."

logging/bin/deploy_esexporter.sh
cumcke: I have gone ahead and updated it in the files that I've touched.
Do you want me to go ahead and update it to DEBUG in **deploy_elasticsearch_content_open.sh** and **deploy_kibana_content.sh** files?

logging/bin/deploy_fluentbit_open.sh
cumcke: I went ahead and updated this to DEBUG to align with the other messages that reference a specific file in the USER_DIR directory.

logging/bin/deploy_esexporter.sh
gsmith-sas: I believe I've already done that as part of my current Pull Request; so I think we are good.

monitoring/bin/deploy_monitoring_cluster.sh
cumcke: I think we should only use the "..." if we have a message that confirms that the task is done.  In the example above, I'll go ahead and remove it, but for messages like the one below, I think we should keep it.

Deploy X ...
X has been deployed

What do you guys think?

BryanEllington: Consistency is the most important thing. Either way is fine as long as there's a guide as to when '...' is used.

gsmith-sas (approved):
LGTM

PR #224, opened 2021-11-12
doc updates for new Kibana release

technomerlyn:


logging/README.md
```diff
@@ -257,6 +257,48 @@ The script creates the namespace into which the components are deployed. By defa
 Updates in place are supported. To update, re-run the
 `deploy_logging_open.sh` script to install the latest versions of all components, indexes, and dashboards.
 
+## Important Information about Kibana in the New Release
+
```
gsmith-sas, line 261: I think we need a better section title.  This document is the top-level README and will be here forever (more or less).  Given that, the phrase "new release" isn't meaningful.  Maybe something like:  "Using Kibana"?

logging/README.md
```diff
@@ -257,6 +257,48 @@ The script creates the namespace into which the components are deployed. By defa
...
+### Overview
+
+**Notes:**
```
gsmith-sas, line 264: I think it is either NOTES or OVERVIEW, but not both.  I like the bullet list presentation and would prefer to keep it rather than present that information in paragraph form.

logging/README.md
```diff
@@ -257,6 +257,48 @@ The script creates the namespace into which the components are deployed. By defa
...
+* As of release 1.1.0, this project now uses Open Distro for Elasticsearch version 1.13.2, which includes updated versions of Elasticsearch and Kibana.
+* In this release, the multi-tenancy capabilities of Kibana have been enabled. This change requires that users select a Kibana tenant space after logging in. For the new procedure, see [Log Into Kibana](#login_kibana).
+* Kibana no longer provides a persistent navigation bar by default. Instead, a menu button ( ![menu button](../img/kibana_menu.png) ) that opens the navigation menu is now provided. You can choose to make the navigation menu persist on the page by clicking __Dock navigation__.
+
+### <a name="login_kibana"></a>Log In to Kibana
+
+Use the following steps to log in to Kibana:
+
+1. In the "Login to Kibana" window, enter your Kibana credentials and click __Log In__. The "Select your tenant" window appears.
+2. In the __Choose from custom__ list, confirm that your tenant is selected by default. If not, select your tenant from the list.
```
gsmith-sas, line 275: I wonder if we should use special formatting for the window titles...to make them stand-out more.  Maybe bold + italic (to differentiate them from the menu text which we have used bold for).

gsmith-sas (commented):
This looks pretty good...but (of course) I have a few comments. 😉 🤷‍♂️   The images were loaded properly and the links work as expected.

logging/README.md
technomerlyn: The title is meaningful at this time, catches attention in a sea of instructions, and gives weight to make sure users do not ignore the information. We can adjust the title as time goes by. Indeed, the topic itself will likely change significantly as time goes by.

logging/README.md
technomerlyn: This section and its formatting follow SAS Style Guidelines.

logging/README.md
technomerlyn: Introducing more typeface formatting is poor usability. Again, the formatting follows SAS Style Guidelines.

PR #238, opened 2021-11-23
"Batch" up Kibana content before importing

gsmith-sas:


logging/bin/import_kibana_content.sh
```diff
@@ -90,6 +113,9 @@ fi
 #Flag to suppress file/directory not found error
 ignore_not_found="${IGNORE_NOT_FOUND:-false}"
 
+#Flag to force loading of directory files individually
+nobatch="${NOBATCH:-false}"
```
BryanEllington, line 117: Instead of the double-negative 'nobatch=false', how about batch=true?

The env var could also be more descriptive. Assume we'll get rid of the double-negative, maybe use BATCH_KIBANA_CONTENT, KIBANA_CONTENT_BATCH_LOAD, or something along those lines?

gsmith-sas: This flag is really meant as an internal flag that I wouldn't expect customers to ever use except in rare circumstances when working with Tech Support.  But, that said, it's probably better for it to err on the side of clarity: I've renamed it to BATCH_KIBANA_CONTENT.

BryanEllington (approved):
Looks good

BryanEllington (approved):
Thumbs up

PR #240, opened 2021-11-24
V4m version tracking

cumcke:
- Updated **deployV4MInfo** and **removeV4MInfo** to take two parameters - a namespace and a release name.
- Updated all invocations of **deployV4MInfo** and **removeV4MInfo** to take two parameters.
- For monitoring, the release names are as follows:  v4m (cluster-level), v4m-viya (SAS Viya level), viya-tenant-$VIYA_TENANT (tenant-level).  For logging, v4m is the only name used (same behavior).
- Created the **getHelmReleaseVersion** function to set the Helm version of a component, given a namespace and release name.

### DIFF
--- a/bin/version-include.sh
+++ b/bin/version-include.sh
@@ -65,6 +65,8 @@ function deployV4MInfo() {
     -n "$NS" \
     --values $valuesYAML \
     $releaseName ./v4m-chart
+
+  getHelmReleaseVersion "$NS" "$releaseName"
 }
 
 function removeV4MInfo() {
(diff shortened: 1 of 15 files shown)


gsmith-sas (commented):
See comments for a suggestion that eliminates the need to touch a good chunk of these files...and a bit of nitpicking related to variable names.  

ALSO: It would be helpful to add a label or annotation on the Kubernetes objects created by this that would allow us to identify the "logging" or "monitoring" namespace.   While those are the default (and, most likely, most common) names for those two namespaces, users are free to choose others.  Having a label or annotation would allow our scripts to identify these namespaces (by finding instances of these objects and checking the label/annotation).  That would be useful.  For example, it would allow a script running in the "logging" namespace to identify the "monitoring" namespace (or vice-versa) so it could create an Elasticsearch datasource for that Grafana instance.  This might be slightly out-of-scope of this Pull Request, but...

bin/version-include.sh
```diff
@@ -63,50 +64,82 @@ function deployV4MInfo() {
...
 function removeV4MInfo() {
   NS=$1
+  releaseName=$2
   if [ -z "$NS" ]; then
     log_error "No namespace specified for removing Viya Monitoring for Kubernetes version information"
     return 1
   fi
   log_info "Removing Viya Monitoring for Kubernetes version information"
-  helm uninstall -n "$NS" v4m
+  helm uninstall -n "$NS" "$releaseName"
```
gsmith-sas, line 80: We should confirm the release exists.  If we don't, this helm command will return a non-zero exit code...and errexit (set -e) will exit.

bin/version-include.sh
```diff
@@ -63,50 +64,82 @@ function deployV4MInfo() {
...
 }
 
-if [ -z "$V4M_VERSION_INCLUDE" ]; then
+function getHelmReleaseVersion() {
```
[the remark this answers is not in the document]
cumcke: The original intention of this function was to get the v4m version information (originally called getV4MInfo or something along those lines); however, with Bryan's help, we were able to expand its functionality to be able to get the deployed version of any component that was deployed with a Helm chart.
Although, we are currently only using it with regards to v4m, it's not limited to just that (which is why it has the more "generic" name.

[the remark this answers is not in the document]
gsmith-sas: Cool.  In that case, we might want to consider moving the function to the helm-include.sh file with the other generic Helm functions.

bin/version-include.sh
```diff
@@ -51,6 +51,7 @@ function populateValuesYAML() {
 
 function deployV4MInfo() {
   NS=$1
+  releaseName=$2
```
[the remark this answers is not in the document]
cumcke: I've added logic to default releaseName to v4m if one is not present.  We'll see it in my next push.

bin/version-include.sh
cumcke: Although it seemed to continue working despite producing an error, I have I added logic to check if the namespace/releaseName combination exists before attempting to do perform the uninstall.  While the code I implemented does work as I expect it to, it keeps giving me a message saying that there are too many square brackets that I'm trying to work through.

bin/version-include.sh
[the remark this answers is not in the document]
cumcke: Got it.  I went ahead and moved it to the helm-include.sh file instead.

bin/version-include.sh
gsmith-sas: It looks like the helm list command will return an empty set of square brackets if the specified release doesn't exist and you've requested yaml output.  
```
-bash-4.2$ helm list -n logging --filter "^does_not_exist$" -o yaml
[]
```
That might be part of what you are running into.

bin/version-include.sh
```diff
@@ -63,50 +64,82 @@ function deployV4MInfo() {
...
-    v4m ./v4m-chart
+    $releaseName ./v4m-chart
+
+  getHelmReleaseVersion "$NS" "$releaseName"
 }
 
 function removeV4MInfo() {
   NS=$1
+  releaseName=$2
   if [ -z "$NS" ]; then
```
gsmith-sas, line 75: If the purpose of this function is to remove the v4m Helm release, why is releaseName a parameter?  Wouldn't it ALWAYS be set to "v4m"?

bin/version-include.sh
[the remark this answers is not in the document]
gsmith-sas: If the purpose of this function is to deploy the v4m Helm release, why is releaseName a parameter?  Wouldn't it ALWAYS be set to "v4m"?

bin/version-include.sh
BryanEllington: We have multiple releases - v4m, v4m-viya, and v4m-tenant-[tenant] - The deploy and remove functions used to be hardcoded to just 'v4m' since that's all we had. Now they're generic enough to handle all 3 situations.

bin/version-include.sh
```diff
@@ -63,50 +64,47 @@ function deployV4MInfo() {
...
-        V4M_CURRENT_STATUS=${BASH_REMATCH[1]}
-      fi
-    done
-
-    log_debug "V4M_CURRENT_VERSION_FULL=$V4M_CURRENT_VERSION_FULL"
-    log_debug "V4M_CURRENT_VERSION_MAJOR=$V4M_CURRENT_VERSION_MAJOR"
-    log_debug "V4M_CURRENT_VERSION_MINOR=$V4M_CURRENT_VERSION_MINOR"
-    log_debug "V4M_CURRENT_VERSION_PATCH=$V4M_CURRENT_VERSION_PATCH"
-    log_debug "V4M_CURRENT_STATUS=$V4M_CURRENT_STATUS"
+  getHelmReleaseVersion "$V4M_NS"
```
BryanEllington, line 90: Add the 'v4m' 2nd parameter here

bin/version-include.sh
```diff
@@ -63,50 +64,47 @@ function deployV4MInfo() {
...
+  
+  V4M_CURRENT_VERSION_FULL=releaseVersionFull
+  V4M_CURRENT_VERSION_MAJOR=releaseVersionMajor
+  V4M_CURRENT_VERSION_MINOR=releaseVersionMinor
+  V4M_CURRENT_VERSION_PATCH=releaseVersionPatch
+  V4M_CURRENT_STATUS=releaseStatus
```
BryanEllington, line 96: Need `$` in front of all variable names on the right side here

bin/version-include.sh
```diff
@@ -63,50 +64,47 @@ function deployV4MInfo() {
...
+  log_debug "V4M_CURRENT_STATUS=$V4M_CURRENT_STATUS"
 
-    export V4M_CURRENT_VERSION_FULL V4M_CURRENT_VERSION_MAJOR V4M_CURRENT_VERSION_MINOR V4M_CURRENT_VERSION_PATCH 
-    export V4M_CURRENT_STATUS
-  fi
+  export V4M_CURRENT_VERSION_FULL V4M_CURRENT_VERSION_MAJOR V4M_CURRENT_VERSION_MINOR V4M_CURRENT_VERSION_PATCH 
+  export V4M_CURRENT_STATUS
 
-  export -f deployV4MInfo removeV4MInfo
+  export -f deployV4MInfo removeV4MInfo getHelmReleaseVersion
```
BryanEllington, line 107: Remove this since `getHelmReleaseVersion` is exported by the helm-include.sh script

bin/helm-include.sh
```diff
@@ -89,6 +89,47 @@ function helmRepoAdd {
...
+  log_debug "releaseVersionMajor=$releaseVersionMajor"
+  log_debug "releaseVersionMinor=$releaseVersionMinor"
+  log_debug "releaseVersionPatch=$releaseVersionPatch"
+  log_debug "releaseStatus=$releaseStatus"
+
+  export releaseVersionFull releaseVersionMajor releaseVersionMinor releaseVersionPatch releaseStatus
+}
+
 export HELM_VER_FULL HELM_VER_MAJOR HELM_VER_MINOR HELM_VER_PATCH
 export -f helm2ReleaseExists
```
BryanEllington, line 134: Add a line: `export -f getHelmReleaseVersion`

BryanEllington (changes requested):

bin/version-include.sh
cumcke: I have updated the getHelmReleaseVersion function to default to 'v4m' if one isn't provided (like the other functions) so it doesn't need to be hard-coded:
`releaseName=${2:-'v4m'}`

bin/helm-include.sh
cumcke: I went ahead and added the line.  It should show up in the next pull request.

gsmith-sas (commented):
Looks good to me...but I will defer to Bryan and leave it to him for to approve or not.

BryanEllington (approved):
LGTM

PR #243, opened 2021-11-29
Remove application multi-tenancy feature flag

gsmith-sas:


logging/bin/deploy_kibana_content.sh
```diff
@@ -122,87 +122,67 @@ if [ "$kibanaready" != "TRUE" ]; then
...
-         log_warn "There was an issue importing the cached existing Kibana content into the Kibana tenant space [cluster_admins]. [$response]"
-         log_debug "Failed response details: $(tail -n1 $kb_migrate_response)"
-         #TODO: Exit here?  Display messages as shown?  Add BIG MESSAGE about potential loss of content?
+      rc=$?
+      if [ "$rc" == "0" ]; then
+         log_debug "The Kibana tenant space [admin_tenant] was deleted."
       else
-         log_info "Existing Kibana imported to [cluster_admins] Kibana tenant space. [$response]"
-         log_debug "Import details: $(tail -n1 $kb_migrate_response)"
+         log_debug "Problems were encountered while attempting to delete tenant space [admin_tenant]."
```
BryanEllington, line 156: Is this supposed to be a log_debug message or should it be log_warn?

gsmith-sas: It's a DEBUG level message because users are (almost certainly) unaware that this tenant was created as part of the original ODFE 1.7.0 deployment.  Kibana multi-tenancy was disabled so they would have no way of ever bumping into this tenant.  If we cannot delete the tenant, throwing a WARNing is going to make the user think they need to do something.  While we don't really need/want that tenant around, it shouldn't cause problems...and it isn't worth having someone call Tech Support about.

BryanEllington (approved):
Looks good. (Very) Minor nit - I think we should avoid the use of periods at the end of log messages unless there are multiple sentences

PR #245, opened 2021-12-01
[CHANGE] Support creation/deletion of RBACs and users with access to …

gsmith-sas:
…all log messages


logging/bin/deploy_kibana_content.sh
```diff
@@ -183,6 +183,41 @@ fi
...
+# create the all logs RBACs
+add_notice "**Elasticsearch/Kibana Access Controls**"
+LOGGING_DRIVER=true ./logging/bin/security_create_rbac.sh _all_ _all_
+
+# Create the 'logadm' Kibana user who can access all logs
+LOG_CREATE_LOGADM_USER=${LOG_CREATE_LOGADM_USER:-false}
+if [ "$LOG_CREATE_LOGADM_USER" == "true" ]; then
+
+   if user_exists logadm; then
+      log_warn "A user 'logadm' already exists; leaving that user as-is.  Review its definition in Kibana and update it, or create another user, as needed."
```
BryanEllington, line 196: "Review its definition in Kibana and update it, or create another user, as needed." seems superfluous

logging/bin/security_create_rbac.sh
```diff
@@ -43,50 +45,71 @@ tenant=${2}
...
 
-   NST="${namespace}_${tenant}"
-   INDEX_NST="${namespace}-__${tenant}__"
 else
-   NST="$namespace"
-   INDEX_NST="${namespace}"
-fi
+   # creating namespace or tenant limited RBACs
+   if [ -z "$namespace" ]; then
+     log_error "Required argument NAMESPACE no specified"
```
BryanEllington, line 72: typo - "no" -> "not"

logging/bin/security_create_rbac.sh
```diff
@@ -43,50 +45,71 @@ tenant=${2}
...
+      INDEX_NST="${namespace}-__${tenant}__"
+   else
+      NST="$namespace"
+      INDEX_NST="${namespace}"
+   fi
+
+   if [ -n "$tenant" ]; then
+     log_notice "Creating access controls for tenant [$tenant] within namespace [$namespace] [$(date)]"
+   else
+     log_notice "Creating access controls for namespace [$namespace] [$(date)]"
```
BryanEllington, line 100: Why is the date important here?

logging/bin/deploy_kibana_content.sh
```diff
@@ -183,6 +183,41 @@ fi
...
+
+# create the all logs RBACs
+add_notice "**Elasticsearch/Kibana Access Controls**"
+LOGGING_DRIVER=true ./logging/bin/security_create_rbac.sh _all_ _all_
+
+# Create the 'logadm' Kibana user who can access all logs
+LOG_CREATE_LOGADM_USER=${LOG_CREATE_LOGADM_USER:-false}
+if [ "$LOG_CREATE_LOGADM_USER" == "true" ]; then
+
+   if user_exists logadm; then
```
BryanEllington, line 195: Should we allow an override of the 'logadm' username? Could be a separate issue from this if it's too much to do/test right now.

logging/bin/deploy_kibana_content.sh
gsmith-sas: The point of the additional sentence is to point out that while a user with this name already exists, it may or may not be configured the way the user expected.  We wouldn't want the person running the deployment script to interpret the fact that this user (name) already exists as "this user with all of the expected permissions has already been created".

logging/bin/security_create_rbac.sh
gsmith-sas: fixed.

logging/bin/security_create_rbac.sh
gsmith-sas: In all (most?) of the component deployment scripts on the logging side, I've included the date/time in the first and last messages.  This allows the user to correlate running the script with changes in behavior later. And, if they have logs from several runs saved somewhere, helps clarify which log corresponds to which run.

logging/bin/deploy_kibana_content.sh
gsmith-sas: LOL, great minds think alike.  I started implementing that and then decided on holding off on that.  New users can be created at any time by running the user.sh script.  Since we're always creating these access controls, additional users with these rights can be created as-needed.  So, if they really hate the 'logadm' user name, they can set the flag to tell us NOT to create it and then create a comparable user with their preferred username by calling the script.  We may still decide to offer the ability to override our default name for this user at some point, but for now I don't think it's critical capability.

BryanEllington (approved):
Thanks for the answers...thumbs up

PR #259, opened 2021-12-17
Event Router now deployed in logging namespace. 

cumcke:
I have added logic in both the deploy_eventrouter.sh (for upgrade-in-place situations) and in the remove_eventrouter.sh (to appropriately remove Event Router in older deployments).

I currently don't have any logic around the removal from kube-system so any suggestions around that would be helpful.


logging/bin/deploy_eventrouter.sh
```diff
@@ -23,17 +36,24 @@ set -e
 # Enable workload node placement?
 LOG_NODE_PLACEMENT_ENABLE=${LOG_NODE_PLACEMENT_ENABLE:-${NODE_PLACEMENT_ENABLE:-false}}
 
+# If performing an upgrade-in-place, check for/remove Event Router in kube-system namespace (if older than 1.1.3) 
+if [ ! -z "$V4M_CURRENT_VERSION_FULL" ]; then
```
gsmith-sas, line 40: The comment above says this is only done if the version is older than 1.1.3, but the check only tests if V4M_CURRENT_VERSION_FULL is not empty.  Was that variable only added in 1.1.3?

logging/bin/remove_eventrouter.sh
```diff
@@ -10,8 +10,23 @@ this_script=`basename "$0"`
...
+log_debug "Replacing logging namespace for files in [$logDir]"
+  if echo "$OSTYPE" | grep 'darwin' > /dev/null 2>&1; then
+    sed -i '' "s/__LOG_NS__/$LOG_NS/g" $logDir/eventrouter.yaml
+  else
+    sed -i "s/__LOG_NS__/$LOG_NS/g" $logDir/eventrouter.yaml
+  fi
+
 log_info "Removing Event Router [$(date)]"
-kubectl delete --ignore-not-found -f logging/eventrouter.yaml
+# Remove existing instance of Event Router in the kube-system namespace (if present).
```
gsmith-sas, line 27: Is this being done to handle "older" deployments?  Or, do we think we will have clusters where there will be two instances of Event Router running?  I wonder if should only delete one instance (with the one in the $LOG_NS being our first/preferred target)?  If our "new" one is there, what makes us think the one in kube-system is "ours" too?  I'm not pushing for anything, just raising the question.

logging/eventrouter/eventrouter.yaml
```diff
@@ -0,0 +1,71 @@
...
+metadata:
+  name: v4m-eventrouter
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: eventrouter
+subjects:
+- kind: ServiceAccount
+  name: eventrouter
+  namespace: "logging"
```
gsmith-sas, line 27: Shouldn't "logging" be "__LOG_NS__" here?

logging/eventrouter/eventrouter.yaml
```diff
@@ -0,0 +1,71 @@
...
+apiVersion: v1
+data:
+  config.json: |-
+    {
+      "sink": "stdout"
+    }
+kind: ConfigMap
+metadata:
+  name: v4m-eventrouter-cm
+  namespace: "logging"
```
gsmith-sas, line 38: See above.

logging/eventrouter/eventrouter.yaml
```diff
@@ -0,0 +1,71 @@
...
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: v4m-eventrouter
+  namespace: "logging"
```
gsmith-sas, line 44: See above.

monitoring/monitors/kube/podMonitor-eventrouter.yaml
```diff
@@ -1,15 +1,15 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
-  name: eventrouter
+  name: v4m-eventrouter
   labels:
```
gsmith-sas, line 5: In an UIP scenrario, do we need to worry about removing the existing podMonitors tied to the previous instance of Event Router (i,.e. the one in kube-system)?

gsmith-sas (changes requested):

logging/eventrouter/eventrouter.yaml
cumcke: I originally hard-coded logging and was going to string replacement using **sed**.  The next pull request will address this.

logging/bin/remove_eventrouter.sh
cumcke: This is currently how I am handling the "older" deployments until I can come up with a more concrete way to identify "our" Event Router.

I haven't implemented the logic needed to stop once after an instance has been found and removed (checking from in $LOG_NS > in kube-systems) ... though, even that could unnecessarily remove an instance in kube-system.

I'm open to any suggestions around this.

logging/eventrouter/eventrouter.yaml
```diff
@@ -1,13 +1,13 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: v4m-eventrouter
-  namespace: "logging"
+  name: eventrouter-v4m
```
gsmith-sas, line 4: I lean towards making the fewest changes necessary.  With that in mind, do we really need to change _every_ references of eventrouter (and/or v4m-eventrouter) to eventrouter-v4m.  I  accept that inverting things in the the pod name gets us over the Fluent Bit hurdle, but I'm not sure there's value in extending that non-standard naming to everything related to eventrouter.  Do you feel that the value of keeping things consistent within the eventrouter "bubble" outweighs the value of keeping things consistent within the larger v4m "bubble"?  Please give that some thought and let's discuss. And if there are _technical_ reasons why resource names must match please call those out.  For example, do the ServiceAccount and ClusterRole names have to match each other and/or the pod name?

logging/fb/fluent-bit_config.configmap_open.yaml
```diff
@@ -195,7 +195,7 @@ data:
        Alias events_renames
        Name  modify
        Match kube.*
-       Condition Key_value_equals  PROPS__appname eventrouter
+       Condition Key_value_matches  PROPS__appname eventrouter*
```
gsmith-sas, line 198: Is this change necessary and/or does it it improve things in some way?  If we do not change the Kubernetes label (see previous comment), do we need this change?

gsmith-sas (changes requested):
I'm not requesting any specific changes at this point but I'd like to chat after you've considered the points raised in my comments.

logging/eventrouter/eventrouter.yaml
cumcke: That makes sense.

I just changed it to where I only change the metadata name (and anything that uses it) to be **eventrouter-v4m** and reverted all other instances back to just **eventrouter**.  It should show up on the next update.

logging/fb/fluent-bit_config.configmap_open.yaml
cumcke: I went ahead and reverted this change.  It should show up in the next update.

monitoring/monitors/kube/podMonitor-eventrouter.yaml
cumcke: After our conversations, I don't think it's necessary to make this change.
I'm reverting it for now and will test it.

gsmith-sas (approved):
LGTM

PR #265, opened 2022-01-05
Fix verbose logging

BryanEllington:


bin/log-include.sh
```diff
@@ -22,8 +22,8 @@ else
   fi
 fi
 
-if [ "$LOG_DEBUG_ENABLE" != "true" ]; then
-  exec >/dev/null
+if [ "$LOG_VERBOSE_ENABLE" != "true" ]; then
```
gsmith-sas, line 25: Should we add a comment here explaining why/what we are doing?  I think having a comment like that here originally would have allowed us to catch this mistake during the initial code reviews.

gsmith-sas (commented):
Kind of disappointing we missed this originally.  I suggest we add a comment to clarify (see my comment).

bin/log-include.sh
BryanEllington: Added comment in latest commit

gsmith-sas: Thank you for the comment.  But, sorry, I meant adding a comment to explain line 26 (the exec 1>/dev/null) since the purpose of that is less obvious.

BryanEllington: Added the comment

gsmith-sas (approved):
Looks good.

cumcke (approved):
Looks good to me.

PR #273, opened 2022-01-18
Set KB_TLS_ENABLE based on TLS_ENABLE by default

gsmith-sas:


logging/user.env
```diff
@@ -81,7 +81,7 @@
 # will attempt to use cert-manager to generate one on-the-fly.
 # Note that these on-the-fly certs will likely lead to browser alerts
 # warning of possible security risks when accessing Kibana.
-# LOG_KB_TLS_ENABLE=false
+# LOG_KB_TLS_ENABLE=$TLS_ENABLE
```
gsmith-sas, line 84: Do we want to mark the LOG_KB_TLS_ENABLE environment variable as "DEPRECATED" in the user.env files?

BryanEllington: yes

BryanEllington (approved):
Looks good

PR #274, opened 2022-01-18
Add 'logadm' to accounts change_internal_password.sh can change

gsmith-sas:


logging/bin/change_internal_password.sh
```diff
@@ -61,56 +65,34 @@ if [ "$NEW_PASSWD" == "" ]; then
...
+      ES_USER=$USER_NAME
+   else
+      ES_USER=$(kubectl -n $LOG_NS get secret internal-user-$USER_NAME -o=jsonpath="{.data.\username}" |base64 --decode)
+      ES_PASSWD=$(kubectl -n $LOG_NS get secret internal-user-$USER_NAME -o=jsonpath="{.data.password}" |base64 --decode)
+   fi
 else
-   ES_USER=$(kubectl -n $LOG_NS get secret internal-user-$USER_NAME -o=jsonpath="{.data.\username}" |base64 --decode)
-   ES_PASSWD=$(kubectl -n $LOG_NS get secret internal-user-$USER_NAME -o=jsonpath="{.data.password}" |base64 --decode)
+  ES_USER=$USER_NAME
+  ES_PASSWD="do_not_know_current_password"
```
BryanEllington, line 81: I assume this is a placeholder. What happens once this is set and why use this value? I don't see another reference to the value.

logging/bin/change_internal_password.sh
```diff
@@ -184,123 +164,107 @@ if [[ $response == 4* ]]; then
...
-          log_notice "| Kibana web-interface.  The 'admin' password should *ONLY* be changed via   |"
-          log_notice "| this script (logging/bin/change_internal_password.sh)                      |"
-          log_notice "+============================================================================+"
+          log_notice "                    *********** IMPORTANT NOTE ***********                   "
+          log_notice "                                                                             "
+          log_notice " After changing the password for the [logcollector] user, you should restart "
+          log_notice " the Fluent Bit pods to ensure log collection is not interrupted.            "
+          log_notice "                                                                             "
+          log_notice " This can be done by submitting the following command:                       "
+          log_notice "         kubectl -n $LOG_NS delete pods -l 'app=fluent-bit, fbout=es'        "
```
BryanEllington, line 238: I thought we were moving away from fixing the width with spaces and trying to center text

logging/bin/change_internal_password.sh
gsmith-sas: ES_PASSWD is the current password for the account and used in the curl command that follows.  That curl uses the user's own credentials to change the password. But, since we don't know the current password for the logadm account, I was providing a dummy value so the curl command wouldn't fail due to a syntax error.  We want the processing to fall through to the use the admin user's credentials to change the logadm user password.  

Using a nonsense value here was the easiest implementation and avoided having to restructure the code more extensively to handle the logadm account.

logging/bin/change_internal_password.sh
gsmith-sas: The notice text was existing code...I removed the box characters that surrounded the notices but didn't bother removing the leading spaces.  I can do that if you think it is worthwhile.

BryanEllington: Probably not worth it as part of this PR. Maybe it's something we can address separately.

BryanEllington (approved):
Alrighty

PR #275, opened 2022-01-18
Adding logic to check for prom-operator before helm uninstall.

cumcke:
It turns out that I added the logic to get rid of the error in December and didn't realize it; however, I found one instance in the remove_monitoring_cluster script that didn't check first.


BryanEllington (approved):
LGTM

PR #276, opened 2022-01-24
Added functionality to automatically add Elasticsearch data source in…

cumcke:
… logging namespace.

The ELASTICSEARCH_DATASOURCE environment variable already existed (only in the deploy_monitoring_cluster script).  

Added logic to count unique namespaces where our Fluent Bit pod exists.  If there is only one, it will set up the Elasticsearch data source using the admin user ID and password.

### DIFF
--- a/monitoring/bin/deploy_monitoring_cluster.sh
+++ b/monitoring/bin/deploy_monitoring_cluster.sh
@@ -65,10 +65,37 @@ fi
 # Elasticsearch Datasource for Grafana
 ELASTICSEARCH_DATASOURCE="${ELASTICSEARCH_DATASOURCE:-false}"
 if [ "$ELASTICSEARCH_DATASOURCE" == "true" ]; then
-  log_verbose "Provisioning Elasticsearch datasource for Grafana"
-  kubectl delete secret -n $MON_NS --ignore-not-found grafana-datasource-es
-  kubectl create secret generic -n $MON_NS grafana-datasource-es --from-file monitoring/grafana-datasource-es.yaml
-  kubectl label secret -n $MON_NS grafana-datasource-es grafana_datasource=1 sas.com/monitoring-base=kube-viya-monitoring
+
+  # Check to see if logging components have already been deployed.
+  if [[ $(kubectl get pods -A -l app.kubernetes.io/instance=v4m-fb -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) == 0 ]]; then
+    log_verbose "No logging components found.  Skipping Elasticsearch data source set up.";
+  elif [[ $(kubectl get pods -A -l app.kubernetes.io/instance=v4m-fb -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) > 1 ]]; then
+    log_verbose "Too many logging components found.  Skipping Elasticsearch data source set up.";
+  else
+    # Replace logging namespace name and password in grafana-datasource-es.yaml file
+    logNS=$(kubectl get pods -A -l app.kubernetes.io/instance=v4m-fb -o custom-columns=:metadata.namespace --no-headers | uniq)
+    monDir=$TMP_DIR/$MON_NS
+    mkdir -p $monDir
+    cp -R monitoring/grafana-datasource-es.yaml $monDir/grafana-datasource-es.yaml
+  
+    # Pull Elasticsearch password in temp directory file
+    adminPass="$(kubectl -n $logNS get secret internal-user-admin -o=jsonpath="{.data.password}" | base64 --decode)"
+    
+    # Replace placeholders
+    log_debug "Replacing logging namespace for files in [$monDir]"
+    if echo "$OSTYPE" | grep 'darwin' > /dev/null 2>&1; then
+      sed -i '' "s/__logNS__/$logNS/g" $monDir/grafana-datasource-es.yaml
+      sed -i '' "s/__adminPass__/$adminPass/g" $monDir/grafana-datasource-es.yaml
+    else
+      sed -i "s/__logNS__/$logNS/g" $monDir/grafana-datasource-es.yaml
+      sed -i "s/__adminPass__/$adminPass/g" $monDir/grafana-datasource-es.yaml
+    fi
+  
+    log_info "Provisioning Elasticsearch datasource for Grafana"
+    kubectl delete secret -n $MON_NS --ignore-not-found grafana-datasource-es
+    kubectl create secret generic -n $MON_NS grafana-datasource-es --from-file $monDir/grafana-datasource-es.yaml
+    kubectl label secret -n $MON_NS grafana-datasource-es grafana_datasource=1 sas.com/monitoring-base=kube-viya-monitoring
+  fi
 else
   log_debug "ELASTICSEARCH_DATASOURCE not set"
   log_debug "Skipping creation of Elasticsearch datasource for Grafana"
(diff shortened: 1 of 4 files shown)


monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -65,10 +65,37 @@ fi
...
-  log_verbose "Provisioning Elasticsearch datasource for Grafana"
-  kubectl delete secret -n $MON_NS --ignore-not-found grafana-datasource-es
-  kubectl create secret generic -n $MON_NS grafana-datasource-es --from-file monitoring/grafana-datasource-es.yaml
-  kubectl label secret -n $MON_NS grafana-datasource-es grafana_datasource=1 sas.com/monitoring-base=kube-viya-monitoring
+
+  # Check to see if logging components have already been deployed.
+  if [[ $(kubectl get pods -A -l app.kubernetes.io/instance=v4m-fb -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) == 0 ]]; then
+    log_verbose "No logging components found.  Skipping Elasticsearch data source set up.";
+  elif [[ $(kubectl get pods -A -l app.kubernetes.io/instance=v4m-fb -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) > 1 ]]; then
+    log_verbose "Too many logging components found.  Skipping Elasticsearch data source set up.";
```
gsmith-sas, line 73: This should never happen....and (I think/hope) should be difficult (if not impossible) to do.  Was this added just to be extra careful or am I wrong about how likely/possible this is?

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -65,10 +65,37 @@ fi
...
+  else
+    # Replace logging namespace name and password in grafana-datasource-es.yaml file
+    logNS=$(kubectl get pods -A -l app.kubernetes.io/instance=v4m-fb -o custom-columns=:metadata.namespace --no-headers | uniq)
```
gsmith-sas, line 76: Hmmm, we use the LOG_NS to identify the "logging" namespace in our deployment scripts.  I wonder if we should use it here as well.  Perhaps the logic should be to use LOG_NS (if set) and then go looking for stuff if it isn't set.  

Also, how comfortable are we that the uniq command will be available?

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -65,10 +65,37 @@ fi
...
+    monDir=$TMP_DIR/$MON_NS
+    mkdir -p $monDir
+    cp -R monitoring/grafana-datasource-es.yaml $monDir/grafana-datasource-es.yaml
+  
+    # Pull Elasticsearch password in temp directory file
+    adminPass="$(kubectl -n $logNS get secret internal-user-admin -o=jsonpath="{.data.password}" | base64 --decode)"
+    
+    # Replace placeholders
+    log_debug "Replacing logging namespace for files in [$monDir]"
+    if echo "$OSTYPE" | grep 'darwin' > /dev/null 2>&1; then
```
gsmith-sas, line 86: I wonder if we should consolidate the OS check and dealing with OS-specific function calls earlier (in common.sh, maybe). Then we could use a single OS-agnostic custom function syntax throughout our code..with the custom function "translating" to generate the OS-specific calls.  Or, we can move to containerization and eliminate the need to worry about the OS variations. 😉

monitoring/grafana-datasource-es.yaml
```diff
@@ -4,8 +4,8 @@ datasources:
 - name: Elastic
   type: elasticsearch
   access: proxy
-  database: kubernetes_cluster-*
-  url: https://v4m-es-client-service.logging:9200
+  database: viya_logs-*
+  url: https://v4m-es-client-service.__logNS__:9200
   basicAuth: true
   basicAuthUser: admin
```
gsmith-sas, line 10: As we talked about, using the "admin" user for now (while this is all experimental) is fine...but eventually we need to move to use a different user: either the logadm user or a custom user created for this purpose.  And, we probably want to look at pulling this out into a stand-alone script so we can do this for the tenant admins  as well.

gsmith-sas (commented):
Looks good.  I did have some thoughts about a few things that I think we should discuss/consider.

monitoring/bin/deploy_monitoring_cluster.sh
BryanEllington: Extra careful...Do you know what happens if you deploy logging to two different namespaces?

monitoring/bin/deploy_monitoring_cluster.sh
BryanEllington: In this case, we can just get the actual namespace instead of asking for it as input. It's either there or not, so I'm not sure of the value of an override here.

To my knowledge, I believe `uniq` is always available.

monitoring/grafana-datasource-es.yaml
BryanEllington: Good point on the tenant admins. We haven't talked about making logs available for the tenant instances of Grafana. That's probably a separate story.

If `logadm` is appropriate here, should we just switch to that?  I think what we really need is a read-only user.

monitoring/bin/deploy_monitoring_cluster.sh
gsmith-sas: Hmmmm, good question.  Just tried it: deployment script fails when it goes to deploy ODFE...Helm sees there are existing resources with the same name and aborts.  Hmmm, we may need/want to add logic to check for a pre-existing deployment in a different namespace to the driver script.  I'll put it on the "to do" list.

monitoring/grafana-datasource-es.yaml
gsmith-sas: Using 'logadm' would be better than using 'admin'.  However, we do not persist the 'logadm' password anywhere so we can automatically get it.  We could create a new user for this purpose.  We could even do that programmatically (we have functions for doing it) in this code or in a stand-alone script version of this code.  A stand-alone script for this could (should) allow for namespace, tenant, user and password to be passed in but also make intelligent default choices for those.

monitoring/bin/deploy_monitoring_cluster.sh
gsmith-sas: I'm less concerned about asking for it but if  LOG_NS has been set in the USER_DIR, it seems we should be using that rather than looking around for the logging components.  I also worry about similarly named variables that serve similar purposes (LOG_NS vs. logNS) causing confusion even if they exist in different scripts.

cumcke: If we assume that both MON_NS and LOG_NS are being appropriately set in their USER_DIR, I can go ahead and explicitly use them.

The only reason why I went with the "search for a pod" approach is because LOG_NS is only declared in the logging/common.sh script (which may not necessarily get called).  In theory, that could be fixed if we start calling logging/common.sh in monitoring scripts, but I wasn't sure if we wanted to go that route (yet).

BryanEllington: I still don't get why we would explicitly use whatever `LOG_NS` is set to when we have to check for the existence of the actual logging deployment there anyway. Why not just get the real namespace instead of setting ourselves up for a mismatch situation?

gsmith-sas: For now, let's leave this as-is.  However, if/when we move to a stand-alone script, we should think about how we would handle the situation where the USER_DIR includes setting LOG_NS to a namespace but we do not find an instance of the logging components deployed there.  It seems like something is amiss in that case and the user should be asked to resolve the discrepancy.  If we have settings that can be interpreted as "global settings" (e.g. LOG_NS, MON_NS), it seems like we should honor them.

gsmith-sas (approved):
LGTM.  We can handle the more elaborate use-cases (not using 'admin' user, supporting multi-tenancy, etc.) in a separate PR if/when we decide to do it.

PR #279, opened 2022-01-26
Created multiline feature flag and added logic to fluentbit deploy script

cumcke:
Added an environment variable to the Fluent Bit configuration file and added logic to the deploy_fluentbit_open.sh script to set the variable appropriately before deploying.

### DIFF
--- a/logging/bin/deploy_fluentbit_open.sh
+++ b/logging/bin/deploy_fluentbit_open.sh
@@ -96,10 +96,11 @@ fi
 log_debug "Using FB ConfigMap:" $FB_CONFIGMAP
 
 # Multiline parser setup
+LOG_MULTILINE_ENABLED=${LOG_MULTILINE_ENABLED}
 if [ "$LOG_MULTILINE_ENABLED" == "true" ]; then
-  LOG_MULTILINE_PARSER="docker, crio"
+  LOG_MULTILINE_PARSER="docker, cri"
 else
-  LOG_MULTILINE_PARSER=
+  LOG_MULTILINE_PARSER=""
 fi
 
 # Create ConfigMap containing Fluent Bit configuration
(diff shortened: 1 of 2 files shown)


logging/bin/deploy_fluentbit_open.sh
```diff
@@ -90,9 +90,17 @@ if [ -f "$USER_DIR/logging/fluent-bit_config.configmap_open.yaml" ]; then
...
    # use copy in repo
    FB_CONFIGMAP="logging/fb/fluent-bit_config.configmap_open.yaml"
+
+   
 fi
 log_debug "Using FB ConfigMap:" $FB_CONFIGMAP
 
+# Multiline parser setup
+if [ "$MULTILINE_ENABLED" == "true" ]; then
+  MULTILINE_ENABLED="docker, crio"
```
gsmith-sas, line 100: Let's rename the environment variable end-users will use to LOG_MULTILINE_ENABLED to be consistent with some of the other environment variables. I'm not sure using the same variable in both places here is a good idea because I think it confuses things that in one context it is a true/false and in another it takes more meaningful values (e.g. "docker,crio").  

There are two ways of handling that: a) create two vars: a boolean used by end-users to enable/disable the feature and a 2nd variable that we pass to the pods with our preferred parsers; or, b) use a single variable and interpret a null value as the equivalent of a setting the flag to "false".    With the 2nd option, the IF statement becomes `if [ -n "$LOG_MULTILINE_PARSER" ]; then` (I would suggest changing the var name in this case to eliminate the "_ENABLED" which suggests a boolean.  A plus (and/pr minus) of the 2nd approach is that it allows (risks?) users providing values other than our preferred "docker, crio" which could cause problems.

gsmith-sas (changes requested):
See comments re: variable naming and possible alternatives.

logging/bin/deploy_fluentbit_open.sh
cumcke: I went with your first approach and created another variable to handle the explicit parsers being used.

gsmith-sas (approved):

PR #283, opened 2022-01-31
Create 1.1.4 Release

gsmith-sas:


BryanEllington (changes requested):
MustFix: The kibana image tag still says 1.13.2

In addition, there are several places that need updating from 1.13.2 to 1.13.3. Some are comments and others are in log output. Some are in older entries in the CHANGELOG, which is fine, but other references should be updated.

Run `grep -R '1\.13\.' .` on the root of the repo to find them.

BryanEllington (approved):
LGTM

PR #284, opened 2022-01-31
Added CPU/memory limit to logging components per Helm chart defaults.

cumcke:
Verified that every pod (except for Event Router) had a CPU/memory request with kubectl get pod --output=yaml -n logging and grepped for requests to confirm that they were all added.
The Event Router default YAML did not specify any limits so I did not add any at this time.


gsmith-sas (commented):
I think the comments on the ODFE helm chart values yaml file are incorrect...other than it looks good.  See comments.

gsmith-sas (approved):
LGTM

PR #288, opened 2022-02-08
Add user workload monitoring enablement for ARO to OpenShift doc

BryanEllington:


monitoring/OpenShift.md
```diff
@@ -87,6 +87,21 @@ oc login [cluster-hostname] -u [userID]
 monitoring/bin/deploy_monitoring_openshift.sh
 ```
 
+(EXPERIMENTAL) - On Azure Red Hat OpenShift, it is necessary to manually
```
gsmith-sas, line 90: Rather than just including this as text within the same section, I think it would be better to have this section (on ARO) appear as a sub-section of the current topic.  So, add a preface header:

### _Azure Red Hat OpenShift (Experimental)_
_On Azure Red Hat Openshift..._

I think that will make it stand out (for ARO users) and make it more clear.

gsmith-sas (commented):
Small formatting change suggested

monitoring/OpenShift.md
BryanEllington: Good idea. I reworked it. Check it out.

Issue #285, opened 2022-02-03
Issues executing this script on MacOS


brdodd:
I'm experiencing issues executing this script in MacOS. The issue seems be related to the sleep function. Specifically, MacOS doesn't like the 's' quantifier associated with the sleep value. When I remove the 's' from all values in each of the scripts, kibana/grafana deploys without issue.

Here are further details:
MacOS 12.0.1
Helm client version: 3.7.1
Kubernetes client version: v1.21.0
Kubernetes server version: v1.20.7
I checked out Viya Monitoring for Kubernetes v1.1.4

BryanEllington:
Seems simple enough. I use https://formulae.brew.sh/formula/coreutils and its `sleep` command supports either duration format.

That said, if the fix is that simple for the built-in MacOS `sleep` command, I'm sure we can get it fixed.

PR #291, opened 2022-02-10
Remove 's' suffix from sleep commands

BryanEllington:


BryanEllington:
This fix is to address #285

PR #300, opened 2022-02-17
Update monitoring component versions

BryanEllington:


gsmith-sas (approved):
I can't/didn't validate that these are, in fact, the "right" version numbers...but the code changes appear reasonable and the various version numbers appear to be consistent.  So, I think you are good to go.

PR #302, opened 2022-02-18
[FIX] Remove hard-coded references to namespace 'logging'

gsmith-sas:


cumcke (approved):
Looks good!

BryanEllington (approved):
I checked as well and these look like the only instances outside of some READMEs of hard-coded namespaces. Looks good.

PR #305, opened 2022-02-22
[Fix] ERR trap emits ERROR message when errexit disabled

gsmith-sas:


bin/common.sh
```diff
@@ -23,14 +23,19 @@ function trap_add() {
       )" "${signal}"
    done
 }
+
 function errexit_msg {
-   log_error "Exiting script due to an error."
+   if [ -o errexit ]; then
+      log_error "Exiting script [`basename $0`] due to an error executing the command [$BASH_COMMAND]."
+   else
+      log_debug "Trap [ERR] triggered in [`basename $0`] while executing the command [$BASH_COMMAND]."
```
BryanEllington, line 31: Is 'ERR' the name of the trap? I'm not sure what that's adding to the message.

Also, is the exit code available? Can we log that?

gsmith-sas: ERR is the name of the trap.  We could leave it out but we do have multiple traps defined (well, 2: one on ERR and one on EXIT), so it seemed like we should include it.  

I just tested whether we still have access to the return code of the failing command and, unfortunately, we do not.  Checking "$?"  in the trap returned "0" even though the failing command (in my test scripts) exits with a non-zero exit code.

BryanEllington: OK

BryanEllington (approved):
Issues answered. LGTM.

cumcke (approved):
Code change looks good to me.

bin/common.sh
```diff
@@ -23,14 +23,19 @@ function trap_add() {
       )" "${signal}"
    done
 }
+
 function errexit_msg {
-   log_error "Exiting script due to an error."
+   if [ -o errexit ]; then
```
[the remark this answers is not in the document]
gsmith-sas: Yes, and yes (your interpretation is correct).

PR #309, opened 2022-03-07
Fix grafana datasource URL when using TLS and path-based ingress

BryanEllington:


samples/generic-base/monitoring/user.env
```diff
@@ -9,6 +9,10 @@
 # This setting effectively sets TLS_ENABLE only for monitoring
 # MON_TLS_ENABLE=false
 
+# Set this value to 'true' if using path-based ingress and
+# TLS_ENABLE=true or MON_TLS_ENABLE=true
+MON_TLS_PATH_INGRESS=false
+
```
gsmith-sas, line 15: Shouldn't this flag be added to the TLS samples too?  And possibly to the ingress samples as well?

gsmith-sas (commented):
I had some questions and I think some other samples need to be updated...see below.

PR #314, opened 2022-03-15
Ifspace

gsmith-sas:


logging/bin/deploy_elasticsearch_open.sh
```diff
@@ -357,7 +357,7 @@ do
 done
 
 if [ -z "$es_status" ];  then
-      log_warn "Unable to verify that the Elasticsearch pod [v4m-es-master-0] has initialized."
+      log_debug "Unable to verify that the Elasticsearch pod [v4m-es-master-0] has initialized."
```
gsmith-sas, line 360: I suggest we change this from a WARN to a DEBUG message since it seems to have been triggered in the bulk of my test cases.  I suspect that does not mean things are any worse than they were, we just weren't checking on things after our previous wait.  But we can discuss.

cumcke (approved):
Looks good

PR #316, opened 2022-03-16
Created RBACs and users for use by Grafana datasource

gsmith-sas:


logging/bin/gfuser.sh
```diff
@@ -0,0 +1,59 @@
+#! /bin/bash
```
gsmith-sas, line 1: This script is only a testing convienence, it will NOT become permanent part of the project.  The logic included will eventually be incorporated into the create_elasticsearch_datasource*.sh script(s).

cumcke (approved):
Code looks good.
Ran through the logic and the expected users and roles were created.

PR #317, opened 2022-03-17
Update releaseName in deploy and remove scripts to specify monitoring/logging

cumcke:
- Updated all of the deploy and remove scripts to set **releaseName** to v4m-monitoring-* or v4m-logging.

### DIFF
--- a/logging/bin/deploy_logging_open.sh
+++ b/logging/bin/deploy_logging_open.sh
@@ -86,7 +86,13 @@ set -e
 # Version Info                   #
 ##################################
 
-if ! deployV4MInfo "$LOG_NS" "v4m-logging"; then
+# If a deployment with the old name exists, remove it first
+if helm3ReleaseExists v4m $LOG_NS; then
+  log_verbose "Removing outdated instance of SAS Viya Monitoring"
+  helm uninstall -n "$LOG_NS" "v4m"
+fi
+
+if ! deployV4MInfo "$LOG_NS" "v4m-log"; then
   log_warn "Unable to update SAS Viya Monitoring version information"
 fi
 
--- a/logging/bin/deploy_logging_open.sh
+++ b/logging/bin/deploy_logging_open.sh
@@ -86,7 +86,7 @@ set -e
 # Version Info                   #
 ##################################
 
-if ! deployV4MInfo "$LOG_NS"; then
+if ! deployV4MInfo "$LOG_NS" "v4m-logging"; then
   log_warn "Unable to update SAS Viya Monitoring version information"
 fi
 
--- a/logging/bin/deploy_logging_open.sh
+++ b/logging/bin/deploy_logging_open.sh
@@ -88,11 +88,11 @@ set -e
 
 # If a deployment with the old name exists, remove it first
 if helm3ReleaseExists v4m $LOG_NS; then
-  log_verbose "Removing outdated instance of SAS Viya Monitoring"
+  log_verbose "Removing outdated SAS Viya Monitoring Helm chart release from [$LOG_NS] namespace"
   helm uninstall -n "$LOG_NS" "v4m"
 fi
 
-if ! deployV4MInfo "$LOG_NS" "v4m-log"; then
+if ! deployV4MInfo "$LOG_NS" "v4m-logs"; then
   log_warn "Unable to update SAS Viya Monitoring version information"
 fi
 
--- a/logging/bin/deploy_logging_open.sh
+++ b/logging/bin/deploy_logging_open.sh
@@ -93,7 +93,7 @@ if helm3ReleaseExists v4m $LOG_NS; then
 fi
 
 if ! deployV4MInfo "$LOG_NS" "v4m-logs"; then
-  log_warn "Unable to update SAS Viya Monitoring version information"
+  log_warn "Unable to update SAS Viya Monitoring Helm chart release"
 fi
 
 # Write any "notices" to console
--- a/logging/bin/deploy_logging_open_openshift.sh
+++ b/logging/bin/deploy_logging_open_openshift.sh
@@ -127,7 +127,7 @@ logging/bin/deploy_servicemonitors_open_openshift.sh
 # Version Info                   #
 ##################################
 log_info "STEP 7: Updating version info"
-if ! deployV4MInfo "$LOG_NS" "v4m"; then
+if ! deployV4MInfo "$LOG_NS" "v4m-logging"; then
   log_warn "Unable to update SAS Viya Monitoring version info"
 fi
 
--- a/logging/bin/deploy_logging_open_openshift.sh
+++ b/logging/bin/deploy_logging_open_openshift.sh
@@ -130,11 +130,11 @@ log_info "STEP 7: Updating version info"
 
 # If a deployment with the old name exists, remove it first
 if helm3ReleaseExists v4m $LOG_NS; then
-  log_verbose "Removing outdated instance of SAS Viya Monitoring"
+  log_verbose "Removing outdated SAS Viya Monitoring Helm chart release from [$LOG_NS] namespace"
   helm uninstall -n "$LOG_NS" "v4m"
 fi
 
-if ! deployV4MInfo "$LOG_NS" "v4m-log"; then
+if ! deployV4MInfo "$LOG_NS" "v4m-logs"; then
   log_warn "Unable to update SAS Viya Monitoring version info"
 fi
 
--- a/logging/bin/deploy_logging_open_openshift.sh
+++ b/logging/bin/deploy_logging_open_openshift.sh
@@ -127,7 +127,14 @@ logging/bin/deploy_servicemonitors_open_openshift.sh
 # Version Info                   #
 ##################################
 log_info "STEP 7: Updating version info"
-if ! deployV4MInfo "$LOG_NS" "v4m-logging"; then
+
+# If a deployment with the old name exists, remove it first
+if helm3ReleaseExists v4m $LOG_NS; then
+  log_verbose "Removing outdated instance of SAS Viya Monitoring"
+  helm uninstall -n "$LOG_NS" "v4m"
+fi
+
+if ! deployV4MInfo "$LOG_NS" "v4m-log"; then
   log_warn "Unable to update SAS Viya Monitoring version info"
 fi
 
--- a/logging/bin/remove_logging_open.sh
+++ b/logging/bin/remove_logging_open.sh
@@ -50,7 +50,9 @@ if [ "$LOG_DELETE_CONFIGMAPS_ON_REMOVE" == "true" ]; then
   kubectl delete configmap --ignore-not-found -n $LOG_NS -l managed-by=v4m-es-script
 fi
 
+# Check for and remove any v4m deployments with old naming convention
 removeV4MInfo "$LOG_NS" "v4m"
+
 removeV4MInfo "$LOG_NS" "v4m-logs"
 
 if [ "$LOG_DELETE_NAMESPACE_ON_REMOVE" == "true" ]; then
--- a/logging/bin/remove_logging_open.sh
+++ b/logging/bin/remove_logging_open.sh
@@ -50,7 +50,13 @@ if [ "$LOG_DELETE_CONFIGMAPS_ON_REMOVE" == "true" ]; then
   kubectl delete configmap --ignore-not-found -n $LOG_NS -l managed-by=v4m-es-script
 fi
 
-removeV4MInfo "$LOG_NS" "v4m-logging"
+# If a deployment with the old name exists, remove it first
+if helm3ReleaseExists "v4m" $LOG_NS; then
+  removeV4MInfo "$LOG_NS" "v4m"
+else
+  removeV4MInfo "$LOG_NS" "v4m-log"
+fi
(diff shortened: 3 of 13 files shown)


monitoring/bin/deploy_monitoring_viya.sh
```diff
@@ -89,5 +89,5 @@ else
   log_warn "Prometheus Operator not found. Skipping deployment of ServiceMonitors."
 fi
 
-deployV4MInfo "$VIYA_NS" "v4m-viya"
+deployV4MInfo "$VIYA_NS" "v4m-monitoring-$VIYA_NS"
```
gsmith-sas, line 92: I'm wondering if we really want to lose the  string "viya" here.  Without it, we can only infer that that this Helm release is a tenant-level one because there is only one string after the "v4m-".  But what if a namespace name includes a hyphen (e.g. kube-system)?  Is the Helm release v4m-kube-system a namespace-level release (for a namespace called 'kube-system') or is it a tenant-level release (for a tenant 'system' within the 'kube' namespace)?  

When constructing the ES index names, we used underscores and plus signs to separate namespace names and tenant names, since neither of those characters was acceptable within either namespace or tenant names.  I'm not pushing to duplicate that here necessarily, just raising the general issue.  Are there limitations on which characters can be included in a Helm release name?

On the other hand, Helm releases are namespace-scoped, so maybe that helps to some degree.  Maybe until someone deploys Viya into a namespace called "logging" (or "monitoring"). 😉

logging/bin/deploy_logging_open.sh
```diff
@@ -86,7 +86,7 @@ set -e
 # Version Info                   #
 ##################################
 
-if ! deployV4MInfo "$LOG_NS"; then
+if ! deployV4MInfo "$LOG_NS" "v4m-logging"; then
```
gsmith-sas, line 89: I'm going to suggest we use "logs" rather than "logging".  Similarly, let's use "metrics" instead of "monitoring" and "traces" when we add support for distributed tracing.

monitoring/bin/deploy_monitoring_tenant_openshift.sh
```diff
@@ -258,7 +258,7 @@ if [ ! "$OPENSHIFT_AUTH_ENABLE" == "true" ]; then
   fi
 fi
 
-deployV4MInfo "$VIYA_NS" "v4m-tenant-$VIYA_TENANT"
```
gsmith-sas, line 261: See note below raising question about naming things for the namespace and tenant levels.

gsmith-sas (changes requested):
Overall, this looks good.  It is a testament to our good design that none of this required any actual code changes (other than to parameters passed to the existing functions). 

 As you can see, I have some thoughts about naming.

 Since we have been shipping, do we need to add logic somewhere to deal with deployments where the previous logic was being used?

logging/bin/deploy_logging_open.sh
```diff
@@ -86,7 +86,13 @@ set -e
...
 ##################################
 
-if ! deployV4MInfo "$LOG_NS"; then
+# If a deployment with the old name exists, remove it first
+if helm3ReleaseExists v4m $LOG_NS; then
+  log_verbose "Removing outdated instance of SAS Viya Monitoring"
+  helm uninstall -n "$LOG_NS" "v4m"
+fi
+
+if ! deployV4MInfo "$LOG_NS" "v4m-log"; then
```
gsmith-sas, line 95: Let's go with "v4m-logs" rather than "v4m-log"...make it plural as you did with the metrics one...(and we can do "traces" if/when we add tracing).

logging/bin/remove_logging_open.sh
```diff
@@ -50,7 +50,13 @@ if [ "$LOG_DELETE_CONFIGMAPS_ON_REMOVE" == "true" ]; then
   kubectl delete configmap --ignore-not-found -n $LOG_NS -l managed-by=v4m-es-script
 fi
 
-removeV4MInfo "$LOG_NS"
+# If a deployment with the old name exists, remove it first
+if helm3ReleaseExists "v4m" $LOG_NS; then
+  removeV4MInfo "$LOG_NS" "v4m"
+else
```
gsmith-sas, line 56: I don't think we want the removal of the "v4m-logs" to be conditional here.  Regardless of whether was (or was not) a "v4m" release, we want to remove "v4m-logs".

How does the removeV4MInfo function handle things if the specified release does NOT exist?  If it handles that without problems (e.g. failing w/o ERRORs) do we need to check if the "v4m" release exists on line 54.  If it does NOT handle that situation well, do we need to add a check that the "v4m-logs" release exists before we attempt to delete it?

logging/bin/deploy_logging_open_openshift.sh
```diff
@@ -127,7 +127,14 @@ logging/bin/deploy_servicemonitors_open_openshift.sh
 # Version Info                   #
 ##################################
 log_info "STEP 7: Updating version info"
-if ! deployV4MInfo "$LOG_NS" "v4m"; then
+
+# If a deployment with the old name exists, remove it first
+if helm3ReleaseExists v4m $LOG_NS; then
+  log_verbose "Removing outdated instance of SAS Viya Monitoring"
```
gsmith-sas, line 133: How about changing the message to "Removing outdated SAS Viya Monitoring Helm chart release from [$LOG_NS] namespace"?

gsmith-sas (changes requested):
A couple of things/suggestions that need to be implemented multiple times:

- let's go with "v4m-logs" rather than "v4m-log"
- We should refer to the Helm release consistently across all of the message.  We currently refer to it as "[an] instance of SAS Viya Monitoring" in some places and "SAS Viya Monitoring version info" in others.  I suggested "SAS Viya Monitoring Helm chart release" in one of my comments but I'm not wed to that.  But whatever we decide to use, we should use it in all of messages"
- I think we probably should include namespace in the messages as well.  Probably should tweak the message to indicate "viya" or "tenant" in someway as well, when appropriate.
- See my note about whether removing the current releases should be conditional or not.  
- And my question about whether it is necessary to check for the existence of a Helm release before deleting it.

logging/bin/remove_logging_open.sh
cumcke: I was originally worried about there being multiple uninstall error messages like the ones below, but mine doesn't attempt it unless it finds an instance first.  I removed the conditional logic and it should show up in my next push:

> Removing logging components from the [test] namespace [Mon Mar 21 15:59:29 EDT 2022]
>
> INFO Removing Fluent Bit components [v4m-fb] from the [test] namespace [Mon Mar 21 15:59:30 EDT 2022]
> **Error: uninstall: Release not loaded: v4m-fb: release: not found**
> 
> INFO Removing Elasticsearch Exporter [Mon Mar 21 15:59:31 EDT 2022]
> **Error: uninstall: Release not loaded: es-exporter: release: not found**
> 
> INFO Removing Open Distro for Elasticsearch [Mon Mar 21 15:59:31 EDT 2022]
> **Error: uninstall: Release not loaded: odfe: release: not found**

logging/bin/deploy_logging_open.sh
cumcke: Updated the names to v4m-logs.  It should show up in the next push.

logging/bin/deploy_logging_open_openshift.sh
cumcke: Updated the message per your recommendation for all deployment scripts.  It should show up in the next push.

monitoring/bin/deploy_monitoring_tenant.sh
```diff
@@ -195,7 +195,7 @@ deploy_tenant_dashboards monitoring/multitenant/dashboards
 
 # If a deployment with the old name exists, remove it first
 if helm3ReleaseExists "v4m-tenant-$VIYA_TENANT" $MON_NS; then
-  log_verbose "Removing outdated instance of SAS Viya Monitoring"
+  log_verbose "Removing outdated [${VIYA_NS}/${VIYA_TENANT}] SAS Viya Monitoring Helm chart release"
```
gsmith-sas, line 198: I believe the message with come out looking like:
_**Removing outdated [d12345/acme] SAS Viya Monitoring Helm chart release**_
I think reworking it a bit might make it more readable, perhaps something like:
**_Removing outdated SAS Viya Monitoring Helm chart release for tenant [d12345/acme]_**

monitoring/bin/deploy_monitoring_tenant_openshift.sh
```diff
@@ -260,7 +260,7 @@ fi
 
 # If a deployment with the old name exists, remove it first
 if helm3ReleaseExists "v4m-tenant-$VIYA_TENANT" $MON_NS; then
-  log_verbose "Removing outdated instance of SAS Viya Monitoring"
+  log_verbose "Removing outdated [${VIYA_NS}/${VIYA_TENANT}] SAS Viya Monitoring Helm chart release"
   helm uninstall -n "$MON_NS" "v4m-tenant-$VIYA_TENANT"
```
gsmith-sas, line 264: see comment above.

monitoring/bin/remove_monitoring_cluster.sh
```diff
@@ -71,12 +71,8 @@ if [ "$MON_DELETE_PVCS_ON_REMOVE" == "true" ]; then
...
 fi
 
-# If a deployment with the old name exists, remove it first
-if helm3ReleaseExists "v4m" $MON_NS; then
-  removeV4MInfo "$MON_NS" "v4m"
-else
-  removeV4MInfo "$MON_NS" "v4m-metrics"
-fi
+removeV4MInfo "$MON_NS" "v4m"
+removeV4MInfo "$MON_NS" "v4m-metrics"
```
gsmith-sas, line 76: See comments above.

monitoring/bin/remove_monitoring_openshift.sh
```diff
@@ -34,12 +34,8 @@ log_info "Removing clusterrole and clusterrolebinding..."
...
 kubectl delete --ignore-not-found clusterrolebinding v4m-grafana-clusterrolebinding
 
-# If a deployment with the old name exists, remove it first
-if helm3ReleaseExists "v4m" $MON_NS; then
-  removeV4MInfo "$MON_NS" "v4m"
-else
-  removeV4MInfo "$MON_NS" "v4m-metrics"
-fi
+removeV4MInfo "$MON_NS" "v4m"
+removeV4MInfo "$MON_NS" "v4m-metrics"
```
gsmith-sas, line 39: See comments above.

monitoring/bin/remove_monitoring_tenant.sh
```diff
@@ -85,11 +85,7 @@ if [ "$OPENSHIFT_CLUSTER" == "true" ]; then
...
 fi
 
-# If a deployment with the old name exists, remove it first
-if helm3ReleaseExists "v4m-tenant-$VIYA_TENANT" $VIYA_NS; then
-  removeV4MInfo "$VIYA_NS" "v4m-tenant-$VIYA_TENANT"
-else
-  removeV4MInfo "$VIYA_NS" "v4m-metrics-${VIYA_TENANT}"
-fi
+removeV4MInfo "$VIYA_NS" "v4m-tenant-$VIYA_TENANT"
+removeV4MInfo "$VIYA_NS" "v4m-metrics-${VIYA_TENANT}"
```
gsmith-sas, line 90: See comment re:comments above

monitoring/bin/remove_monitoring_tenant_openshift.sh
```diff
@@ -75,11 +75,7 @@ for type in "${types[@]}"; do
...
 done
 
-# If a deployment with the old name exists, remove it first
-if helm3ReleaseExists "v4m-tenant-$VIYA_TENANT" $VIYA_NS; then
-  removeV4MInfo "$VIYA_NS" "v4m-tenant-$VIYA_TENANT"
-else
-  removeV4MInfo "$VIYA_NS" "v4m-metrics-${VIYA_TENANT}"
-fi
+removeV4MInfo "$VIYA_NS" "v4m-tenant-$VIYA_TENANT"
+removeV4MInfo "$VIYA_NS" "v4m-metrics-${VIYA_TENANT}"
```
gsmith-sas, line 80: See comment re:comments above

monitoring/bin/remove_monitoring_viya.sh
```diff
@@ -40,11 +40,7 @@ if [ "$MON_DELETE_PVCS_ON_REMOVE" == "true" ]; then
...
 fi
 
-# If a deployment with the old name exists, remove it first
-if helm3ReleaseExists "v4m-viya" $VIYA_NS; then
-  removeV4MInfo "$VIYA_NS" "v4m-viya"
-else
-  removeV4MInfo "$VIYA_NS" "v4m-metrics-viya"
-fi
+removeV4MInfo "$VIYA_NS" "v4m-viya"
+removeV4MInfo "$VIYA_NS" "v4m-metrics-viya"
```
gsmith-sas, line 45: See comment re:comments above

gsmith-sas (changes requested):
Sorry, a few more nits about wording...

gsmith-sas (approved):
Looks good.

PR #327, opened 2022-04-14
Combined create_elasticsearch_datasource scripts and updated to special RBAC users

cumcke:


cumcke:
Once the combined scripts have been approved, I will do one more update to remove the individual create_elasticsearch_datasource_*.sh scripts.

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -0,0 +1,203 @@
...
+
+# Check for parameters - set with cluster or tenant
+if [ -z "$tenantNS" ] && [ -z "$tenant" ]; then
+   cluster="true"
+elif [ -n "$tenant" ]; then
+    nst="${tenantNS}_${tenant}"
+else
+   nst="$tenantNS"
+fi
+
```
gsmith-sas, line 79: So we are supporting the creation of a datasource scoped to a single namespace (without tenant)?  I am NOT opposed to that and think it could be useful...but don't recall talking about that.

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -0,0 +1,203 @@
...
+    if [[ $(kubectl get pods -n $MON_NS -l app.kubernetes.io/instance=v4m-prometheus-operator -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) == 0 ]]; then
+        log_error "No monitoring components found in the [$MON_NS] namespace."
+        log_error "Monitoring needs to be deployed in this namespace in order to configure Elasticsearch with this script.";
+        exit 1
+    else
+        log_debug "Monitoring found in $MON_NS namespace.  Continuing."
+    fi
+else
+    log_info "Checking the [$tenantNS] namespace for monitoring deployment for the [$tenant] tenant ..."
+    if [[ $(kubectl get pods -n $tenantNS -l app.kubernetes.io/instance=v4m-grafana-$tenant -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) == 0 ]]; then
```
gsmith-sas, line 92: Hmmm, I suspect this will fail if there is no tenant.  So, we either need to handle that use-case in some way here or flag it as an error if the user provides a namespace but no tenant.

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -0,0 +1,203 @@
...
+    else
+        log_debug "Monitoring deployment was found for the $tenant tenant in the $tenantNS namespace.  Continuing."
+    fi
+fi
+
+# Check to see if logging namespace provided exists and components have already been deployed
+log_info "Checking for Elasticsearch pods in the $LOG_NS namespace ..."
+if [[ $(kubectl get pods -n $LOG_NS -l app=v4m-es -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) == 0 ]]; then
+  log_error "No logging components found in the $LOG_NS namespace."
+  log_error "Logging needs to be deployed in this namespace in order to configure Elasticsearch with this script.";
```
gsmith-sas, line 105: Reword messages.  The code checks for Elasticsearch, so the message should probably say it was not found.  And, this script does NOT configure _Elasticsearch_, it configures a(n Elasticsearch) datasource in Grafana.  

Perhaps:
`log_error "Elasticsearch was not found in the $LOG_NS namespace."`
`log_error "All of the required log monitoring components need to be deployed in this namespace before this script can configure the requested Grafana datasource "`

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -0,0 +1,203 @@
...
+        log_error "Monitoring needs to be deployed in this namespace in order to configure Elasticsearch with this script.";
+        exit 1
+    else
+        log_debug "Monitoring found in $MON_NS namespace.  Continuing."
+    fi
+else
+    log_info "Checking the [$tenantNS] namespace for monitoring deployment for the [$tenant] tenant ..."
+    if [[ $(kubectl get pods -n $tenantNS -l app.kubernetes.io/instance=v4m-grafana-$tenant -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) == 0 ]]; then
+        log_error "No monitoring components were found for $tenant in the $tenantNS namespace."
+        log_error "Monitoring needs to be deployed in this namespace in order to configure Elasticsearch with this script.";
```
gsmith-sas, line 94: Reword message.  Tell them what needs to be done; e.g. the tenant-level metric monitoring must be deployed in this namespace for this tenant (by running the deploy_monitoring_tenant.sh script) before this script can be run.

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -0,0 +1,203 @@
...
+
+get_sec_api_url
+get_credentials_from_secret admin
+
+# FOR TENANT - Check to see if tenant has been deployed in logging.
+if [ "$cluster" != "true" ]; then
+    log_info "Checking that [$tenant] has been onboarded to Elasticsearch in the $LOG_NS namespace ..."
+
+    if ! kibana_tenant_exists "${tenantNS}_${tenant}"; then
+        log_error "Unable to configure Elasticsearch data source."
```
gsmith-sas, line 123: "datasource" or "data source"  Check the Grafana doc and let's be sure we're consistent.

Reword the two log messages. Perhaps:
`log_error "Unable to configure a Grafana datasource for this tenant because the log monitoring onboarding process has not been completed for the [$tenant] in the [$tenantNS] namespace."`
`log_error "This can be done by running the logging/onboard.sh script"`

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -0,0 +1,203 @@
...
+   log_verbose "Removing the existing [$grfds_user] utility account."
+   delete_user $grfds_user
+fi
+
+grfds_passwd="$(randomPassword)"
+
+if [ "$cluster" == "true" ]; then
+   ./logging/bin/user.sh CREATE -ns _all_ -t _all_ -u $grfds_user -p "$grfds_passwd" -g
+elif [ -z "$tenant" ]; then
+   ./logging/bin/user.sh CREATE -ns $tenantNS -u $grfds_user -p "$grfds_passwd" -g
```
gsmith-sas, line 148: See notes above about whether we're going to support creating datasources limited to a namespace (but not to a specific tenant w/in that namespace).

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -0,0 +1,203 @@
...
+if [ "$cluster" == "true" ]; then
+    kubectl create secret generic -n $MON_NS grafana-datasource-es --from-file $monDir/grafana-datasource-es.yaml
+    kubectl label secret -n $MON_NS grafana-datasource-es grafana_datasource=1 sas.com/monitoring-base=kube-viya-monitoring
+else
+    kubectl create secret generic -n $tenantNS v4m-grafana-datasource-es-$tenant --from-file $monDir/grafana-datasource-es.yaml
+    kubectl label secret -n $tenantNS v4m-grafana-datasource-es-$tenant grafana_datasource-$tenant=true sas.com/monitoring-base=kube-viya-monitoring
+fi
+
+# Delete pods so that they can be restarted with the change.
+log_info "Elasticsearch data source provisioned.  Restarting pods to apply the change"
```
gsmith-sas, line 194: Do we really need to kill the pods to have the newly provisioned datasource available?  Seems like the datasource provisioning sidecar should be able to do what is needed without having us manually deleting the pod.

gsmith-sas (changes requested):
- Need to decide whether we support the scenario where the user provides a namespace but no tenant.  That could be useful but the script currently only partially supports it (e.g. doesn't error out).  I guess in that scenario, the datasource should be defined in the cluster-level Grafana instance.
- Various message wording tweaks 
- Please confirm
  -  "datasource" vs. "data source"
  - whether we need to kill the grafana pods to get the new datasources to show up.

monitoring/bin/create_elasticsearch_datasource.sh
cumcke: While it will eventually fail because of the pod checks, I can add some more specific error handling to make sure a namespace and tenant are provided before doing the kubectl get calls.

monitoring/bin/create_elasticsearch_datasource.sh
cumcke: Unfortunately, we did need to kill the pods in order for the change to take effect.  When I was making the scripts initially, the Grafana UI would not pick up the change unless the pods were restarted.

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -90,22 +90,22 @@ if [ "$cluster" == "true" ]; then
 else
     log_info "Checking the [$tenantNS] namespace for monitoring deployment for the [$tenant] tenant ..."
     if [[ $(kubectl get pods -n $tenantNS -l app.kubernetes.io/instance=v4m-grafana-$tenant -o custom-columns=:metadata.namespace --no-headers | uniq | wc -l) == 0 ]]; then
-        log_error "No monitoring components were found for $tenant in the $tenantNS namespace."
-        log_error "Monitoring needs to be deployed in this namespace in order to configure Elasticsearch with this script.";
+        log_error "No monitoring components were found for [$tenant] in the [$tenantNS] namespace."
+        log_error "Monitoring needs to be deployed using the deploy_monitoring_tenant script in order to configure Elasticsearch with this script.";
         exit 1
```
gsmith-sas, line 95: Replace "Elasticsearch" with "the requested Grafana datasource".

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -181,7 +181,7 @@ else
 fi
 
 # Adds the Elasticsearch data source to Grafana
-log_info "Provisioning Elasticsearch datasource for Grafana"
+log_info "Provisioning Elasticsearch data source for Grafana"
```
gsmith-sas, line 184: Drop "Elasticsearch".  Isn't really needed and eliminates a maintenance task when we move to OpenSearch.

gsmith-sas (changes requested):
I think the logic changes are fine but noticed some obsolete/inaccurate messages.

monitoring/bin/create_elasticsearch_datasource.sh
```diff
@@ -191,13 +193,13 @@ else
 fi
 
 # Delete pods so that they can be restarted with the change.
-log_info "Elasticsearch data source provisioned.  Restarting pods to apply the change"
+log_info "Elasticsearch ata source provisioned in Grafana.  Restarting pods to apply the change"
 if [ "$cluster" == "true" ]; then
```
gsmith-sas, line 197: "ata" probably should be "data"

monitoring/bin/create_elasticsearch_datasource.sh
cumcke: Since this is the create_elasticsearch_datasource script, I ended up tweaking the verbiage to reference the "Elasticesearch data source in Grafana" instead.
Since the script would require multiple changes, we will create a separate ticket for Opensearch efforts.

monitoring/bin/create_elasticsearch_datasource.sh
gsmith-sas: I think this fell through the cracks and has not been addressed.

gsmith-sas (changes requested):

gsmith-sas (approved):
Looks good.  Don't forget to remove the previous scripts before merging.

PR #336, opened 2022-04-22
Fixed pod name output when multiple tenants are deployed

cumcke:


gsmith-sas (approved):
Looks pretty straightforward...approved.

PR #339, opened 2022-04-26
Updated Openshift compatibility to accept 4.7+

cumcke:
Changed the Openshift compatibility to accept all versions of Openshift 4 from 4.7 onward.


gsmith-sas (approved):
Looks good.

PR #340, opened 2022-05-02
Updated deploy scripts to use new Elasticsearch datasource script

cumcke:
Updated the following scripts to use the new create_elasticsearch_datasource.sh scripts:

- monitoring/bin/deploy_monitoring_cluster.sh
- monitoring/bin/deploy_monitoring_tenant.sh

### DIFF
--- a/monitoring/bin/deploy_monitoring_cluster.sh
+++ b/monitoring/bin/deploy_monitoring_cluster.sh
@@ -62,15 +62,6 @@ else
   log_debug "Skipping deployment of federated scrape of Istio Prometheus instance"
 fi
 
-# Elasticsearch Datasource for Grafana
-ELASTICSEARCH_DATASOURCE="${ELASTICSEARCH_DATASOURCE:-false}"
-if [ "$ELASTICSEARCH_DATASOURCE" == "true" ]; then
-  configureElasticsearchDatasource
-else
-  log_debug "ELASTICSEARCH_DATASOURCE not set"
-  log_debug "Skipping creation of Elasticsearch datasource for Grafana"
-fi
-
 # Check if Prometheus Operator CRDs are already installed
 PROM_OPERATOR_CRD_UPDATE=${PROM_OPERATOR_CRD_UPDATE:-true}
 PROM_OPERATOR_CRD_VERSION=${PROM_OPERATOR_CRD_VERSION:-v0.54.0}
@@ -227,6 +218,24 @@ if [ $? == 0 ]; then
   kubectl patch prometheusrule --type='json' -n $MON_NS v4m-kubernetes-apps --patch "$(cat monitoring/kube-hpa-alert-patch.json)"
 else
   log_debug "PrometheusRule $MON_NS/v4m-kubernetes-apps does not exist"
+fi
+
+# Elasticsearch Datasource for Grafana
+# Moved down to make sure Grafana pods exist
+ELASTICSEARCH_DATASOURCE="${ELASTICSEARCH_DATASOURCE:-false}"
+if [ "$ELASTICSEARCH_DATASOURCE" == "true" ]; then
+  set +e
+  monitoring/bin/create_elasticsearch_datasource.sh
+
+  if (( $? == 1 )); then
+    log_warn "Unable to configure the Elasticsearch data source at this time."
+    log_warn "Please address the errors and re-run the follow command:"
+    log_warn "monitoring/bin/create_elasticsearch_datasource.sh"
+  fi
+  set -e
+else
+  log_debug "ELASTICSEARCH_DATASOURCE not set"
+  log_debug "Skipping creation of Elasticsearch datasource for Grafana"
 fi
 
 echo ""
--- a/monitoring/bin/deploy_monitoring_cluster.sh
+++ b/monitoring/bin/deploy_monitoring_cluster.sh
@@ -225,11 +225,12 @@ fi
 ELASTICSEARCH_DATASOURCE="${ELASTICSEARCH_DATASOURCE:-false}"
 if [ "$ELASTICSEARCH_DATASOURCE" == "true" ]; then
   set +e
+  log_debug "Creating Elasticsearch datasource using the create_elasticesearch_datasource script"
   monitoring/bin/create_elasticsearch_datasource.sh
 
   if (( $? == 1 )); then
     log_warn "Unable to configure the Elasticsearch data source at this time."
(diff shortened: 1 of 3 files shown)


monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -229,6 +220,24 @@ else
   log_debug "PrometheusRule $MON_NS/v4m-kubernetes-apps does not exist"
 fi
 
+# Elasticsearch Datasource for Grafana
+# Moved down to make sure Grafana pods exist
+ELASTICSEARCH_DATASOURCE="${ELASTICSEARCH_DATASOURCE:-false}"
+if [ "$ELASTICSEARCH_DATASOURCE" == "true" ]; then
+  set +e
+  monitoring/bin/create_elasticsearch_datasource.sh
```
gsmith-sas, line 228: I see no messages indicating successful deployment of the datasource.  Is that because the script itself will emit messages indicating success?

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -229,6 +220,24 @@ else
...
+
+  if (( $? == 1 )); then
+    log_warn "Unable to configure the Elasticsearch data source at this time."
+    log_warn "Please address the errors and re-run the follow command:"
```
gsmith-sas, line 232: I'm not comfortable with a WARNing message talking about "errors"...if there were errors, shouldn't it be an ERROR message?  But I agree that a failure to create the datasource should NOT cause an ERROR here.  So, let's tweak the wording of the message.  It would be nice if the called script had a flag that told it to fail silently or, at least, without an ERROR message.  But that's not there and isn't absolutely needed. Maybe reword things as "You may configure the datasource later, after addressing any issues noted above, by running the following command:"

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -229,6 +220,24 @@ else
...
+    log_warn "monitoring/bin/create_elasticsearch_datasource.sh"
+  fi
+  set -e
+else
+  log_debug "ELASTICSEARCH_DATASOURCE not set"
+  log_debug "Skipping creation of Elasticsearch datasource for Grafana"
```
gsmith-sas, line 238: Let's refer to the datasource as "Viya Logs datasource for Grafana" in this script (and eventually use that elsewhere as well).  That eliminates the (soon to be inaccurate) reference to Elasticsearch and is probably more meaningful to users anyway.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -229,6 +220,24 @@ else
   log_debug "PrometheusRule $MON_NS/v4m-kubernetes-apps does not exist"
 fi
 
+# Elasticsearch Datasource for Grafana
+# Moved down to make sure Grafana pods exist
```
gsmith-sas, line 224: This looks like a message to code reviewers rather a comment for future developers.  It's only meaningful if you know what was here before.  There's no reason to remove it or reword it (necessarily) but remember who the audience for the comments are.

monitoring/bin/deploy_monitoring_tenant.sh
```diff
@@ -170,6 +170,23 @@ helm upgrade --install $helmDebug \
...
+  if (( $? == 1 )); then
+    log_warn "Unable to configure the Elasticsearch data source at this time."
+    log_warn "Please address the errors and re-run the follow command:"
+    log_warn "monitoring/bin/create_elasticsearch_datasource.sh -ns ${VIYA_NS} -t ${VIYA_TENANT}"
+  fi
+  set -e
+else
+  log_debug "ELASTICSEARCH_DATASOURCE not set"
+  log_debug "Skipping creation of Elasticsearch datasource for Grafana"
+fi
```
gsmith-sas, line 188: All notes from above probably apply here as well.

gsmith-sas (changes requested):
Just some tweaks to messages...but looks good otherwise.

monitoring/bin/deploy_monitoring_cluster.sh
cumcke: There is a message in the create_elasticsearch_datasource.sh script that does it, but I'll add a DEBUG or INFO message in the deploy monitoring scripts to explicitly say that it's trying to create this data source.  It will be added in the next pull request.

monitoring/bin/deploy_monitoring_cluster.sh
cumcke: So the problem here is that I'm not aware of a way to make it choose a log level for the error depending on whether it's run independently or not.
When run on its own, it IS supposed to error out.  When run within the deployment scripts, it should still show the regular error (so that people know how to fix it), but the WARNING assures that the script will still run.  I think it's important that, although an error popped up, they can re-run the script and try again after it has been addressed.

monitoring/bin/deploy_monitoring_cluster.sh
cumcke: While I didn't add this line of code, I disagree with changing the wording of this particular message.

The environment variable is ELASTICSEARCH_DATASOURCE.  The script is create_ELASTICSEARCH_datasource.  When we fully convert over to OpenSearch, we can discuss how we want to approach the two data source system; however, a new environment variable and script will be needed and we can adjust/replace the message then.

gsmith-sas: We have several environment variables that reference Elasticsearch or Kibana in some way.  We're going to have to live with that for some period of time.  We will (probably) eventually move to new names and support both during some transition phase.  We're trying to address references to those product names in messages and file names as we move to OpenSearch.  This is a DEBUG level message, so I suppose changing this now isn't critical.

monitoring/bin/deploy_monitoring_cluster.sh
gsmith-sas: Ok, so there currently isn't a mechanism that would allow us to have create_elasticsearch_datasource.sh generate a WARNing rather than an ERROR message when it fails.  While that would be nice to have, it isn't critical.  The current wording here makes the ERROR messages seem more significant than they are.  And, tells the user they need to do something.  The reality is the ERRORs aren't a big deal (which is why we are continuing) and the user only has to do something if they want to make the log messages available in Grafana.  I believe my suggested rewording conveys that information.  If you feel it needs to be tweaked, feel free.

gsmith-sas (approved):
Looks good to me.

PR #344, opened 2022-05-10
Logging dashboards now deploy when logging data source is added.

cumcke:
Added logic to the create_elasticsearch_datasource.sh script to deploy the logging-related dashboards.

Since I wasn't able to get the "deploy single dashboard" logic to work from within another script, we're going to call the script and set the flag for the other dashboards to false (so that only the logging dashboards are deployed).


gsmith-sas (approved):
LGTM

PR #346, opened 2022-05-11
Updated Viya log dashboards to use new Elasticsearch data source name

cumcke:
- The create_elasticsearch_datasource script now deploys the Viya-log dashboards (CAS/Go/Java Services) instead of the logging ones (Elasticsearch/Fluent Bit)
- Updated the CAS, Go, and Java Services dashboards to use the new Elasticsearch data source name.


gsmith-sas (approved):
Looks good.

PR #349, opened 2022-05-19
Move to OpenSearch: File renames and deletions

gsmith-sas:


cumcke (approved):
Looks good from what I can see!

PR #356, opened 2022-06-03
Support OpenSSL for cert generation

gsmith-sas:


cumcke (approved):
Looks good from what I can see.

PR #357, opened 2022-06-03
Updated Postgres and Postgres Database dashboards to include working namespace filter

cumcke:
- Updated "Postgres" dashboard's namespace variable to look for **namespace** instead of **kubernetes-namespace**.
- Updated "Postgres Database" dashboard to include a **namespace** variable.

NOTE:  Although the files may look significantly different, all that was done was the change to the namespace variable and then saving the resulting dashboard as a JSON file.


gsmith-sas (approved):
Hmmm, Git diffs aren't very useful here...so I can't really see the changes.  But approving.

PR #367, opened 2022-06-20
Configurenp

gsmith-sas:


bin/configure_nodeport.sh
```diff
@@ -0,0 +1,113 @@
...
+      servicename="v4m-alertmanager"
+      target_port="${target_port:-31091}"
+      ;;
+   "PROMETHEUS"|"PROM"|"PRO"|"PR")
+      namespace=$MON_NS
+      appname="Prometheus"
+      servicename="v4m-prometheus"
+      target_port="${target_port:-31090}"
+      ;;
+   "GRAFANA"|"GRAF"|"GR")
```
cumcke, line 82: I noticed that there are references in the monitoring deployment scripts that are related to setting up the NodePort.  Should we leave them in as-is or should we use a call to this script instead?

Example from **deploy_monitoring_cluster**:

nodePortValuesFile=$TMP_DIR/empty.yaml
PROM_NODEPORT_ENABLE=${PROM_NODEPORT_ENABLE:-true}
if [ "$PROM_NODEPORT_ENABLE" == "true" ]; then
  log_debug "Enabling NodePort access for Prometheus and Alertmanager"
  nodePortValuesFile=monitoring/values-prom-nodeport.yaml
fi

cumcke (approved):
Questions answered.  Code looks good.

PR #369, opened 2022-06-22
New script to change Grafana admin password and v4m-grafana secret

cumcke:
Created a new script (change_grafana_admin_password.sh) that:

- updates the Grafana admin password using the Grafana CLI
- Updates the v4m-grafana secret with the new admin password so that they match

### DIFF
--- a/monitoring/bin/change_grafana_admin_password.sh
+++ b/monitoring/bin/change_grafana_admin_password.sh
@@ -1,0 +1,133 @@
+#! /bin/bash
+
+# Copyright © 2022, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+cd "$(dirname $BASH_SOURCE)/../.."
+source monitoring/bin/common.sh
+
+this_script=`basename "$0"`
+
+function show_usage {
+  log_message  "Usage: $this_script [--namespace NAMESPACE --tenant TENANT]"
+  log_message  ""
+  log_message  "Changes the password of the Grafana admin user."
+  log_message  ""
+  log_message  "To change the Grafana admin user at the cluster level, do not pass any "
+  log_message  "arguments.  To change the Grafana admin user at the tenant level, you need"
+  log_message  "to provide the following arguments:"
+  log_message  "     -ns, --namespace NAMESPACE   - The namespace where the Viya tenant resides."
+  log_message  "     -t,  --tenant TENANT         - The tenant whose logging data source you want to set up."
+  log_message  "     -p,  --password PASSWORD         - The tenant whose logging data source you want to set up."
+  log_message  ""
+}
+
+# Assigning passed in parameters as variables for the script:
+POS_PARMS=""
+
+# Setting passed in variables:
+while (( "$#" )); do
+  case "$1" in
+    -ns|--namespace)
+      if [ -n "$2" ] && [ ${2:0:1} != "-" ]; then
+        tenantNS=$2
+        shift 2
+      else
+        log_error "A value for parameter [NAMESPACE] has not been provided." >&2
+        show_usage
+        exit 2
+      fi
+      ;;
+    -t|--tenant)
+      if [ -n "$2" ] && [ ${2:0:1} != "-" ]; then
+        tenant=$2
+        shift 2
+      else
+        log_error "A value for parameter [TENANT] has not been provided." >&2
+        show_usage
+        exit 2
+      fi
+      ;;
+    -p|--password)
+      if [ -n "$2" ] && [ ${2:0:1} != "-" ]; then
+        password=$2
+        shift 2
+      else
+        log_error "A value for parameter [PASSWORD] has not been provided." >&2
+        show_usage
+        exit 2
+      fi
+      ;;
+    -h|--help)
+      show_usage
+      exit
+      ;;
+    -*|--*=) # unsupported flags
+      log_error "Unsupported flag $1" >&2
+      show_usage
+      exit 1
+      ;;
+    *) # preserve positional arguments
+      POS_PARMS="$POS_PARMS $1"
+      shift
+      ;;
+  esac
+done
+
+# Convert namespace and tenant to all lower-case
+tenantNS=$(echo "$tenantNS"| tr '[:upper:]' '[:lower:]')
+tenant=$(echo "$tenant"| tr '[:upper:]' '[:lower:]')
+
+# Check for parameters - set with cluster or tenant
+if [ -z "$tenantNS" ] && [ -z "$tenant" ]; then
+  cluster="true"
+  namespace=$MON_NS
+  v4mSecret="v4m-grafana"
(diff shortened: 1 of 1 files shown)


monitoring/bin/change_grafana_admin_password.sh
```diff
@@ -0,0 +1,133 @@
...
+fi
+
+# Check and make sure that a password was provided.
+if [ -z $password ]; then
+  log_error "A value for parameter [PASSWORD] has not been provided"
+  exit 1
+fi 
+
+if [ "$cluster" == "true" ]; then
+  grafanaPod="$(kubectl get pods -n $MON_NS -l app.kubernetes.io/name=grafana --template='{{range .items}}{{.metadata.name}}{{end}}')"
```
gsmith-sas, line 102: The reference to $MON_NS here is (technically) correct, but based on line 84 above, we probably should refer to $namespace here (since we took the time to set it).  It's not a big deal but if we need to change something later, this would force us to hunt for references to both.

monitoring/bin/change_grafana_admin_password.sh
```diff
@@ -0,0 +1,133 @@
...
+
+# Patch new password in Kubernetes
+encryptedPassword="$(echo -n "$password" | base64)"
+log_info "Updating Grafana secret with the new password"
+kubectl -n $namespace patch secret $v4mSecret --type='json' -p="[{'op' : 'replace' ,'path' : '/data/admin-password' ,'value' : '$encryptedPassword'}]"
+
+# Restart Grafana pods and wait for them to restart
+log_info "Restarting the Grafana pods"
+kubectl delete pods -n $namespace -l "app.kubernetes.io/instance=v4m-prometheus-operator" -l "app.kubernetes.io/name=grafana"
+kubectl -n $namespace wait pods --selector "app.kubernetes.io/instance=v4m-prometheus-operator","app.kubernetes.io/name=grafana" --for condition=Ready --timeout=2m
```
gsmith-sas, line 132: Why the wait?  Is this just to make sure the pod is restarted before the user sees the message reporting success?  What happens if the wait fails?  Should you be checking for that and providing an ERROR or WARNING message?

monitoring/bin/change_grafana_admin_password.sh
```diff
@@ -0,0 +1,133 @@
...
+  exit 1
+fi
+
+# Changes the admin password using the Grafana CLI
+log_info "Updating the admin password using Grafana CLI"
+kubectl exec -n $namespace $grafanaPod -c grafana -- bin/grafana-cli admin reset-admin-password $password
+
+# Exit out of the script if the Grafana CLI call fails
+if (( $? == 1 )); then
+  log_error "An error occurred when updating the password"
```
gsmith-sas, line 120: Does a failure always return a 1?  Or is any non-zero return code a sign of a problem?  Might want to use a check that the exit code is NOT 0.

monitoring/bin/change_grafana_admin_password.sh
```diff
@@ -0,0 +1,133 @@
...
+  exit 1
+fi
+
+# Patch new password in Kubernetes
+encryptedPassword="$(echo -n "$password" | base64)"
+log_info "Updating Grafana secret with the new password"
+kubectl -n $namespace patch secret $v4mSecret --type='json' -p="[{'op' : 'replace' ,'path' : '/data/admin-password' ,'value' : '$encryptedPassword'}]"
```
gsmith-sas, line 127: Do we need to (should we) confirm the secret exists? Either earlier in the script or prior to patching it?  That should never be an issue, but...

monitoring/bin/change_grafana_admin_password.sh
```diff
@@ -0,0 +1,133 @@
...
+function show_usage {
+  log_message  "Usage: $this_script [--namespace NAMESPACE --tenant TENANT]"
+  log_message  ""
+  log_message  "Changes the password of the Grafana admin user."
+  log_message  ""
+  log_message  "To change the Grafana admin user at the cluster level, do not pass any "
+  log_message  "arguments.  To change the Grafana admin user at the tenant level, you need"
+  log_message  "to provide the following arguments:"
+  log_message  "     -ns, --namespace NAMESPACE   - The namespace where the Viya tenant resides."
+  log_message  "     -t,  --tenant TENANT         - The tenant whose logging data source you want to set up."
```
gsmith-sas, line 20: Does this script set up "logging data sources"?  I suspect a cut-and-paste without updating the text completely.

gsmith-sas (changes requested):
Looks pretty good...I did find a couple of things needing tweaks.

monitoring/bin/change_grafana_admin_password.sh
cumcke: It was copy-paste that wasn't updated properly.  It will be updated in the next push.

monitoring/bin/change_grafana_admin_password.sh
cumcke: I'm not sure how else it could return so I'll switch it to NOT 0 like you suggested.

monitoring/bin/change_grafana_admin_password.sh
cumcke: I have made a habit of doing a wait when I restart pods just to make sure that the pods are in a "usable" state by the time the script is done.  I included a similar wait in the create_logging_datasource script, as well.

monitoring/bin/change_grafana_admin_password.sh
```diff
@@ -116,18 +116,24 @@ log_info "Updating the admin password using Grafana CLI"
...
 
 # Restart Grafana pods and wait for them to restart
-log_info "Restarting the Grafana pods"
-kubectl delete pods -n $namespace -l "app.kubernetes.io/instance=v4m-prometheus-operator" -l "app.kubernetes.io/name=grafana"
-kubectl -n $namespace wait pods --selector "app.kubernetes.io/instance=v4m-prometheus-operator","app.kubernetes.io/name=grafana" --for condition=Ready --timeout=2m
-log_info "Grafana admin password has been updated"
+log_info "Grafana admin password has been updated.  Restarting Grafana pods to apply the change"
+if [ "$cluster" == "true" ]; then
+    kubectl delete pods -n $MON_NS -l "app.kubernetes.io/instance=v4m-prometheus-operator" -l "app.kubernetes.io/name=grafana"
+    kubectl -n $MON_NS wait pods --selector "app.kubernetes.io/instance=v4m-prometheus-operator","app.kubernetes.io/name=grafana" --for condition=Ready --timeout=2m
```
gsmith-sas, line 133: Hmmm, referring to $MON_NS here again...and then to $tenantNS in line 137...can't you just use $namespace here (just like the change you just made above)?

cumcke: That was a goof-up, too.  It'll be addressed in the next push.

gsmith-sas (approved):
Ok, looks good.

PR #370, opened 2022-06-23
OpenSSL Support: Add labels identifying cert-generator and v4m

gsmith-sas:


cumcke (approved):
Simple enough addition.  LGTM.

PR #389, opened 2022-08-10
Added sas-viya tag to SAS Launched Jobs dashboards

cumcke:
The SAS Launched Jobs dashboard changes look busy (because I had to update the JSON file using the "Save JSON to File" option in Grafana), but all that was changed with both dashboards was the addition of the sas-viya label.

### DIFF
--- a/monitoring/dashboards/viya/sas-launched-jobs-node.json
+++ b/monitoring/dashboards/viya/sas-launched-jobs-node.json
@@ -3,10 +3,7 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": {
-          "type": "datasource",
-          "uid": "grafana"
-        },
+        "datasource": "-- Grafana --",
         "enable": true,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
@@ -23,17 +20,16 @@
   },
   "editable": true,
   "fiscalYearStartMonth": 0,
+  "gnetId": null,
   "graphTooltip": 0,
-  "iteration": 1660157399432,
+  "id": 94,
+  "iteration": 1637607145730,
   "links": [],
   "liveNow": false,
   "panels": [
     {
       "collapsed": false,
-      "datasource": {
-        "type": "prometheus",
-        "uid": "prometheus"
-      },
+      "datasource": null,
       "gridPos": {
         "h": 1,
         "w": 24,
@@ -42,22 +38,12 @@
       },
       "id": 20,
       "panels": [],
-      "targets": [
-        {
-          "datasource": {
-            "type": "prometheus",
-            "uid": "prometheus"
-          },
-          "refId": "A"
-        }
-      ],
       "title": "Current Active Totals",
       "type": "row"
     },
     {
-      "datasource": {
-        "uid": "$datasource"
-      },
+      "cacheTimeout": null,
+      "datasource": "$datasource",
       "fieldConfig": {
         "defaults": {
           "decimals": 2,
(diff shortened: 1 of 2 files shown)


monitoring/dashboards/viya/sas-launched-jobs-node.json
```diff
@@ -85,9 +99,12 @@
         "text": {},
         "textMode": "value_and_name"
       },
-      "pluginVersion": "8.2.1",
+      "pluginVersion": "9.0.3",
       "targets": [
```
gsmith-sas, line 103: Does this mean you did development on a system running Grafana 9.x?  If so, will these updated dashboards work with our currently shipping configuration using Grafana 8.x ?

monitoring/dashboards/viya/sas-launched-jobs-node.json
```diff
@@ -156,9 +172,12 @@
...
```
gsmith-sas, line 176: See comment above re: Grafana 9.x vs. Grafana 8.x

monitoring/dashboards/viya/sas-launched-jobs-node.json
```diff
@@ -3,7 +3,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
```
gsmith-sas, line 9: Is this change in datasource meaningful?  Does it reflect a Grafana version difference or something else?  I see multiple instances of similar changes to datasource throughout this file.

monitoring/dashboards/viya/sas-launched-jobs-node.json
```diff
@@ -20,16 +23,17 @@
...
+  "iteration": 1660157399432,
   "links": [],
   "liveNow": false,
   "panels": [
     {
       "collapsed": false,
-      "datasource": null,
+      "datasource": {
+        "type": "prometheus",
+        "uid": "prometheus"
```
gsmith-sas, line 35: Datasource changed from "null" to "prometheus"?  Why?

monitoring/dashboards/viya/sas-launched-jobs-node.json
```diff
@@ -1405,14 +1439,25 @@
...
       "type": "row"
     }
   ],
   "refresh": "10s",
-  "schemaVersion": 31,
+  "schemaVersion": 36,
   "style": "dark",
-  "tags": [],
+  "tags": [
+    "sas-viya"
```
gsmith-sas, line 1459: I'm not a Grafana expert, but this is the only change to this file that I would have expected to have been needed.  Couldn't you just hand-edit the existing JSON file to add this?  I don't know if that would work but...

monitoring/dashboards/viya/sas-launched-jobs-users.json
```diff
@@ -1419,7 +1419,9 @@
   "refresh": "10s",
   "schemaVersion": 31,
   "style": "dark",
-  "tags": [],
+  "tags": [
+    "sas-viya"
+  ],
   "templating": {
```
gsmith-sas, line 1425: The fact that there is only one change to this file makes me think we should have been able to achieve the same results on the other file with a single change.

gsmith-sas (changes requested):
I'm concerned why adding a tag to a dashboard results in 311 changes to one case but only 1 in the 2nd case.  I don't know what impact all of those changes had (or didn't have).  The 2nd case lines up more with my expectations of what changes might have been necessary for this kind of change.  

I'm also wondering if some of these changes reflect you working with a Grafana 9.x instance rather than Grafana 8.x that reflects our current configuration.

I don't know that any of the changes are unnecessary but the stark differences between the two files seems like a red flag.

monitoring/dashboards/viya/sas-launched-jobs-node.json
cumcke: I just tested the pluginVersion on Grafana 8.4 and all of the dashboards still worked.

gsmith-sas (approved):
The scope of this set of changes seems more appropriate for the desired outcome.  Approving.

gsmith-sas (approved):
Looks good.

PR #392, opened 2022-08-17
Updating various monitoring components.

cumcke:
Updating the following monitoring components:

- kube-prometheus-stack 32.2.1 -> 36.6.1
- Prometheus 2.33.3 -> 2.36.2
- Prometheus Operator 0.54.0 -> 0.57.0
- Grafana 8.4.1 -> 9.0.3
- AlertManager 0.23.0 -> 0.24.0
- Kube State Metrics 2.3.0 -> 2.5.0
- PushGateway 1.4.2 -> 1.4.3


monitoring/dashboards/pgmonitor/pg-details.json
```diff
@@ -69,7 +69,7 @@
         "alertThreshold": true
       },
       "percentage": false,
-      "pluginVersion": "8.4.1",
```
gsmith-sas, line 72: I wonder if we really needed to touch the dashboard JSON files.  It doesn't look like that was done in previous upgrades to the monitoring components.  I assume it doesn't break anything.  Did you find an issue that updating the dashboards was necessary?

gsmith-sas (commented):
This looks good and appears correct (well, I see a consistent set of numbers being used).  I did have a couple of questions (see below).

monitoring/multitenant/tls/mt-grafana-tls-values.yaml
```diff
@@ -16,7 +16,7 @@ extraContainers: |
     - --key=/cert/tls.key
     - --cert=/cert/tls.crt
     - --disable-authentication
-    image: squareup/ghostunnel:v1.5.2
+    # image: squareup/ghostunnel:v1.5.2
     imagePullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
cumcke: I'm not sure why this was commented out.  It might've been a mishap from when I was checking into the "native TLS" ticket. Will remove it in the next push.

monitoring/dashboards/pgmonitor/pg-details.json
cumcke: I thought that we would need to update the plug-in version with our Grafana version (since the versions numbers matched); however, they were not changed in Bryan's last monitoring update.
I'll revert the **pluginVersion** values in the next push.

gsmith-sas (approved):
LGTM as far as I can tell.

PR #396, opened 2022-08-24
[UPGRADE] Fluent Bit 1.9.6

gsmith-sas:


logging/fb/fluent-bit_config.configmap_opensearch.yaml
```diff
@@ -4,7 +4,7 @@ data:
     [FILTER]
        Name modify
        Match *
-       Set  fb_configMap_version 0.2.18
+       Set  fb_configMap_version 0.2.19_test
```
cumcke, line 7: Should this still be 0.2.19_test or does it need to be switched back to 0.2.19?

cumcke (approved):
If the test part is correct, then I can go ahead and approve.
Otherwise, I approve once it's been updated.

logging/fb/fluent-bit_config.configmap_opensearch.yaml
gsmith-sas: The "_test" is a temporary value we remove during the end-of-sprint clean up prior to publishing a new release.

PR #402, opened 2022-09-20
Preparation for version 1.2.6 release.

cumcke:


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # SAS Viya Monitoring for Kubernetes
 
+## Version 1.2.3 (20SEP22)
+
+* **Overall**
+  * [FEATURE] - The V4M container now includes OpenSSL so that you can run the deployment with TLS.
```
gsmith-sas, line 6: - We had a preferred order for items but I don't recall it right now.  Let's use FEATURE, FIX, UPGRADE.  So, group the two FEATURES together and then show the FIX (NOT! "BUG"). ;-)
- How about we simplify things: [FEATURE] OpenSSL added to V4M container.
- (And, technically, that might be a FIX rather than a FEATURE, but...)

CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
...
+  * [BUG] - Adjusted OpenSSL command that created an unnecssary file that caused issues if multiple users tried running the deployment scripts on the same machine.
```
gsmith-sas, line 7: Sorry, I described the temp file issue in the "logging" section but it should be up here in the Overall.  So, you can move the item under "logging" up to here and drop this one (since it's the same change).

CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
...
+  * [UPGRADE] - Grafana has been upgraded from version 8.4.1 to 9.0.3
+  * [UPGRADE] - AlertManager  has been upgraded from version to 0.24.0
+  * [UPGRADE] - Kube State Metrics has been upgraded from version 2.3.0 to 2.5.0
+  * [UPGRADE] - PushGateway has been upgraded from version 1.4.2 to 1.4.3
+
+* **Logging**
+  * [UPGRADE] - Fluent Bit has been upgraded from version 1.8.7 to 1.9.6
+  * [FIX] - Removed unnecessary retrieval of the OpenSearch 'admin' credentials from logging/bin/deploy_osd.sh
+  * [FIX] - A temporary file created when using OpenSSL for TLS certificate generation is no longer left behind after certs are generated
+
```
gsmith-sas, line 23: See note above...this FIX should be in the OVERALL section and NOT in the Logging section...sorry about that.

gsmith-sas (changes requested):
A couple of minor wording/ordering tweaks.

gsmith-sas (approved):

PR #406, opened 2022-09-22
Upgrade Opensearch to 1.3.5 and ES Exporter to 1.5.0

cumcke:


gsmith-sas (changes requested):
I think we may need/want to update the OpenSearch Helm Chart version in deploy_opensearch.sh as well.  Looks like the helm chart was updated a few times over the last few months as well: https://github.com/opensearch-project/helm-charts/compare/opensearch-1.11.1...opensearch-1.14.1.

gsmith-sas (changes requested):
Need to update OSD_HELM_CHART_VERSION in deploy_osd.sh as well.

PR #409, opened 2022-09-26
Fixed typo in alertmanagerSpec variable in some files

cumcke:
Found a sample user.env file that also contained the typo so it was fixed there, too.


gsmith-sas (approved):
LGTM

PR #410, opened 2022-10-10
add sed command to strip leading whitespace for mac compatability

ceelias:
- [x] Works on Mac

- [x] Test w/ docker container to make sure it still works on linux

### DIFF
--- a/logging/bin/deploy_opensearch.sh
+++ b/logging/bin/deploy_opensearch.sh
@@ -73,13 +73,13 @@ if [ ! -f  $TMP_DIR/es-transport.pem ]; then
    log_debug "Extracting es-transport cert from secret"
    kubectl -n $LOG_NS get secret es-transport-tls-secret -o=jsonpath="{.data.tls\.crt}" |base64 --decode > $TMP_DIR/es-transport.pem
 fi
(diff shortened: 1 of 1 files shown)


logging/bin/deploy_opensearch.sh
```diff
@@ -73,13 +73,13 @@ if [ ! -f  $TMP_DIR/es-transport.pem ]; then
    log_debug "Extracting es-transport cert from secret"
    kubectl -n $LOG_NS get secret es-transport-tls-secret -o=jsonpath="{.data.tls\.crt}" |base64 --decode > $TMP_DIR/es-transport.pem
 fi
-node_dn=$(openssl x509 -subject -nameopt RFC2253 -noout -in $TMP_DIR/es-transport.pem | sed  "s/subject=\s*\(\S*\)/\1/")
+node_dn=$(openssl x509 -subject -nameopt RFC2253 -noout -in $TMP_DIR/es-transport.pem | sed  "s/subject=\s*\(\S*\)/\1/" | sed -e 's/^[ \t]*//')
```
gsmith-sas, line 77: What is the "-e" option on sed doing?

ceelias: Was trying to use "-e" to chain sed commands. WIll clean this up

ceelias: "-e" will let you chain sed expressions, opted for that rather than piping in this case

PR #411, opened 2022-10-10
Fix bug that causes v4m-container to fail to build if KUBECONFIG file not provided.

cumcke:
A bug was discovered where the container would fail to build if **$V4M_PATH/.kube/config** did not exist.
Creating an empty KUBECONFIG file if one isn't provided.

### DIFF
--- a/v4m-container/Dockerfile
+++ b/v4m-container/Dockerfile
@@ -26,8 +26,8 @@ RUN mkdir -p $V4M_PATH/bin $V4M_PATH/.kube $V4M_PATH/user_dir
 WORKDIR $V4M_PATH
 
 # Copy contents from kubeconfig and user_dir to the container
-COPY kubeconfig/* /opt/v4m/.kube/
-COPY user_dir/* /opt/v4m/user_dir/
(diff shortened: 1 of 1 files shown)


v4m-container/Dockerfile
```diff
@@ -25,9 +25,12 @@ ENV KEEP_TMP_DIR=true
 RUN mkdir -p $V4M_PATH/bin $V4M_PATH/.kube $V4M_PATH/user_dir
 WORKDIR $V4M_PATH
 
+# Create $V4M_PATH/.kube/config if it doesn't exist
+RUN touch ${KUBECONFIG}
```
ceelias, line 29: Did you test this change for the case where the config file does exist? 
I'm just wondering if this command creating it, causes issue on line 32 when trying to copy that file over or does it just overwrite them empty version created here?

cumcke: I tested the change locally.

If the file doesn't exist, it creates the file and assigns the empty file those permissions.
If the file does exist, the touch updates the "Last Modified" date.

ceelias (approved):

PR #414, opened 2022-10-12
Fix v4m tag ignored issue in dockerfile

ceelias:


.gitignore
```diff
@@ -3,4 +3,5 @@
 tmp/
 *-tmp.yaml
 user-dir/
-user-dir-*/
\ No newline at end of file
+user-dir-*/
+v4m-container/kubeconfig/
```
ceelias, line 7: Since docs recommend copying `kubeconfig` here, we should just ignore it from git to avoid accidental pushes of configs

v4m-container/Dockerfile
```diff
@@ -1,13 +1,13 @@
...
-ARG V4M_TAG="stable"
-
 ARG ALPINE_TAG=latest
 
 # Need glibc to support the OpenShift 'oc' CLI
 # https://hub.docker.com/r/frolvlad/alpine-glibc/
 FROM frolvlad/alpine-glibc:$ALPINE_TAG
 
+# Version of the Viya Monitoring for Kubernetes deployment to download to the container
+ARG V4M_TAG="stable"
```
ceelias, line 8: This was being ignored because it was before the `FROM`, which essentially resets the ARG, so it was being treated as blank.

v4m-container/Dockerfile
```diff
@@ -35,48 +35,46 @@ ENV TERM=xterm-256color
...
+  ncurses \
+  curl \
+  git \
+  jq \
+  wget \
+  yq \
+  openssh \
+  procps \
+  openssl \
+  ;
```
ceelias, line 51: My editor automatically made these format changes. I can revert this if that is preferred

v4m-container/Dockerfile
```diff
@@ -35,48 +35,46 @@ ENV TERM=xterm-256color
...
 
 # viya4-monitoring-kubernetes
-RUN git clone https://github.com/sassoftware/viya4-monitoring-kubernetes; \
-	cd viya4-monitoring-kubernetes; \
-	V4M_BRANCH="$V4M_TAG"; \
-	if [ "$V4M_BRANCH" == "latest" ]; then V4M_BRANCH="stable"; fi; \
-	git checkout ${V4M_BRANCH}
+RUN V4M_BRANCH="$V4M_TAG"; \
+  if [ "$V4M_BRANCH" == "latest" ]; then V4M_BRANCH="stable"; fi; \
+  git clone --depth 1 -b ${V4M_BRANCH} https://github.com/sassoftware/viya4-monitoring-kubernetes
```
ceelias, line 78: This is a small improvement to only clone the repo and specific branch with the last commit rather than pulling down all git commits

cumcke (approved):
Looks good to me.

PR #418, opened 2022-10-20
Change fluentbit record modifier to use allowlist instead of whitelist

ceelias:


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # SAS Viya Monitoring for Kubernetes
 
+## UNRELEASED
+
+* **Overall**
+
+* **Metrics**
+
+* **Logging**
+  * [FIX] Updated Fluent Bit configuration to use `Allowlist_key` instead of `Whitelist_key` modifier as it has been deprecated and is ignored.
```
ceelias, line 10: I think we should make a process change to start including CHANGELOG changes with our PRs. This will make end of sprint activities a bit easier and just have us update the version at the top

gsmith-sas: I think that is a good idea.  Let's try it out in this sprint and see if merge conflicts (due to CHANGELOG changes) become an issue.

CHANGELOG.md
```diff
@@ -11,8 +20,8 @@
   * [FIX] Customizations in the user-values-prom-operator.yaml file under the alertmanagerSpec will now be applied correctly.
 
 * **Logging**
-  *  [UPGRADE] Opensearch and Opensearch Dashboards has been upgraded from 1.32 to 1.35
-  *  [UPGRADE] Elasticsearch Exporter has been upgraded from 1.2.1 to 1.5.0
+  * [UPGRADE] Opensearch and Opensearch Dashboards has been upgraded from 1.32 to 1.35
+  * [UPGRADE] Elasticsearch Exporter has been upgraded from 1.2.1 to 1.5.0
```
gsmith-sas, line 25: Out of curiosity: was this a human-made change or did the editor do this as part of an automatic cleanup/linter process?

logging/fb/fluent-bit_config.configmap_azmonitor.yaml
```diff
@@ -900,20 +900,20 @@ data:
        Remove_regex   __temp_.+
     [FILTER]
```
gsmith-sas, line 901: When we make changes to the Fluent Bit configuration configMap, we bump the version number (fb_configMap_version) on line 7 and append a string (e.g. "test", or "a").  This makes it easier to confirm the updated configuration is actually being used.  At the end of the sprint, we have a task to remove the additional string and confirm the version number if "correct".  I believe line 7 currently reads: 
`       Set  fb_configMap_version 0.2.19`
So, you might change it to read:
`       Set  fb_configMap_version 0.2.20-test`
It could be argued that the trailing text isn't needed and forces us to remove it in a separate task at the end of the spring, but it is helpful if the file needs to be revised multiple times in a single sprint.

logging/fb/fluent-bit_config.configmap_opensearch.yaml
```diff
@@ -900,20 +900,20 @@ data:
        Remove_regex   __temp_.+
     [FILTER]
```
gsmith-sas, line 901: see note above re: fb_configMap_version value.

logging/fb/fluent-bit_config.configmap_azmonitor.yaml
```diff
@@ -900,20 +900,20 @@ data:
...
        #
-       # WHITELIST Vars to output
+       # ALLOWLIST Vars to output
        #
```
gsmith-sas, line 904: I wonder if we should add a comment like: 
`# Whitelist_key removed as alias for Allowlist_key in FB 1.9.x`
Or do we think the commit message is sufficient documentation?

gsmith-sas (changes requested):
See comments...primary change requested relates to fb_configMap_version field...which you would have no way of knowing about.

CHANGELOG.md
ceelias: Sounds good!

CHANGELOG.md
ceelias: This was actually a manual change, just happened to notice it and made the adjustment

logging/fb/fluent-bit_config.configmap_azmonitor.yaml
gsmith-sas: Ok, I'm with skipping the comment.

gsmith-sas (approved):
Looks good.

PR #420, opened 2022-10-20
set nodeports to be disabled by default for prometheus and alertmanager

ceelias:
- [x] Validated by myself

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * **Overall**
 
 * **Metrics**
-  * [DEPRECATION] NodePorts is now disabled by default for Prometheus and AlertManager for security reasons. Set the environment variable PROM_NODEPORT_ENABLE=true to replicate previous behavior. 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # SAS Viya Monitoring for Kubernetes
 
+## UNRELEASED
+
+* **Overall**
+
+* **Metrics**
+  * [DEPRECATION] NodePorts is now disabled by default for Prometheus and AlertManager for security reasons. Set the environment variable PROM_NODEPORT_ENABLE=true to replicate previous behavior.
```
gsmith-sas, line 8: Would prefer describing the change more explicitly; something like: "For security reasons, access to Prometheus and AlertManager via NodePort is no longer enabled by default.  Set the environment variable PROM_NODEPORT_ENABLE=true to replicate previous behavior."

gsmith-sas (changes requested):
Please tweak CHANGELOG wording but otherwise looks good.

CHANGELOG.md
ceelias: Makes sense, will do!

gsmith-sas (approved):
Thanks.  Looks good.

PR #422, opened 2022-11-03
Updating monitoring components

cumcke:
Performed the following updates:
  * [UPGRADE] - Kube-prometheus-stack has been upgraded from version 36.6.1 to 41.7.3
  * [UPGRADE] - Prometheus has been upgraded from version 2.36.2 to 2.39.0
  * [UPGRADE] - Prometheus Operator has been upgraded from version 0.57.0 to 0.60.0
  * [UPGRADE] - Grafana has been upgraded from version 9.0.3 to 9.2.3
  * [UPGRADE] - K8s-sidecar used with Grafana has been upgraded from 1.19.2 to 1.19.5

Also corrected a statement in the CHANGELOG - the AlertManager was missing its previous version.


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 * **Metrics**
   * [DEPRECATION] For security reasons, access to Prometheus and AlertManager via NodePort is no longer enabled by default. Set the environment variable PROM_NODEPORT_ENABLE=true to replicate previous behavior.
+  * [UPGRADE] - Kube-prometheus-stack has been upgraded from version 36.6.1 to 41.7.3
+  * [UPGRADE] - Prometheus has been upgraded from version 2.36.2 to 2.39.0
+  * [UPGRADE] - Prometheus Operator has been upgraded from version 0.57.0 to 0.60.0
+  * [UPGRADE] - Grafana has been upgraded from version 9.0.3 to 9.2.3
```
gsmith-sas, line 12: Are the versions of the thing we are updating the "default" versions deployed by this new version of the kube-prom-stack Helm chart?  

Are we upgrading Alertmanager, Kube State Metrics and Pushgateway this time?  While we are primarily focused on the one CVE, I think we want to be consistent in how we upgrade things as much as possible.  So, we probably should bump these other components to whatever their default versions are based on the new version of the kube-prom-stack Helm chart.

monitoring/values-prom-operator.yaml
```diff
@@ -62,7 +62,7 @@ kubeStateMetrics:
 # https://github.com/helm/charts/tree/master/stable/kube-state-metrics
 kube-state-metrics:
   image:
-    tag: v2.5.0
+    tag: v2.6.0
   resources:
```
gsmith-sas, line 66: Hmm, looks like we _did_ update Kube State Metrics...need to include that in the CHANGELOG.

gsmith-sas (changes requested):
See comments.  Also, I think we need to add logic to the deployment script to delete that daemonset as mentioned in the kube-prom-stack Helm chart notes about upgrading.  Just add a check if the daemonset exists, if it does, delete it.

CHANGELOG.md
cumcke: I bumped the version of components that happened to have updates.  Kube State Metrics did have an update (and forgot to note it as you saw); however, Alertmanager and the Pushgateway did not have any available updates since the last time I performed the update.
It will be added in the next push.

monitoring/values-prom-operator.yaml
cumcke: Good catch!  I included it in the CHANGELOG for the next push.

gsmith-sas (approved):
Looks good to me...thanks for all the hard work!

PR #448, opened 2023-01-10
Check if tls certs are expired if they already exist

ceelias:


bin/tls-include.sh
```diff
@@ -105,15 +105,22 @@ function create_tls_certs_cm {
...
     TLS_SECRET_NAME=$app-tls-secret
-    if [ -z "$(kubectl get secret -n $namespace $TLS_SECRET_NAME -o name 2>/dev/null)" ]; then
-        if [ "$deployedIssuers" == "false" ]; then
-          deploy_issuers $namespace $context
-          deployedIssuers="true"
-        fi
-        deploy_app_cert "$namespace" "$context" "$app"
-    else
-      log_debug "Using existing $TLS_SECRET_NAME for [$app]"
+    if [ -n "$(kubectl get secret -n $namespace $TLS_SECRET_NAME -o name 2>/dev/null)" ]; then
```
ceelias, line 108: I flipped the logic of this if block so that it would check if the secret does exist and then check if certs are expired or not.

bin/tls-include.sh
```diff
@@ -105,15 +105,22 @@ function create_tls_certs_cm {
...
+       if kubectl get secret -n $namespace $TLS_SECRET_NAME -o "jsonpath={.data['tls\.crt']}" | base64 -d | openssl x509 -enddate -noout -checkend 86400
```
ceelias, line 109: This will pull the secret down, decode the cert, and then use openssl command to check if it is either expired or will expire within the next 24 hours.

1. Any issues with checking within 24 hours, should we configure a different time frame?
2. `openssl` can be used to check the cert here even if it was issued by certmanager. don't think thats an issue to use openssl even though this is the `cert-manager` flow but just wanted to point that out

ceelias:
Couple of things I've considered but want some feedback on:
1. Currently will check if cert is expired or will expire within 24 hours. Should the timeframe be different and should that value be configurable by the user(Not sure how much use that would get)?
2. Should there be a flag that tells the script to not delete and rotate the certs even if they are expired? (This will only happen if TLS_ENABLE=True and the certs already exist)
3. Should there be another entrypoint to allow certs to be rotated without having to run the deploy script?

bin/tls-include.sh
gsmith-sas: The only TLS certs we should replace are those that we generated.  So, if the organization has deployed using their own certs (i.e. not had us generate certs using openssl or cert-manager), we should NOT replace their certs even if they are expired or about to expire.   I believe (but please confirm) that the secrets containing the certs we generate (regardless of cert generation mechanism) have a label of "managed-by" with a value of "v4m" (i.e. "managed-by=v4m").    

In addition, if cert-manager is being used, cert-manager will actually generate new certs when the old certs expire.  The challenge is that these certs are not picked up by the pod automatically, so while we do not have to regenerate the certs, the pods need to be restarted to pick them up.  But I don't know if we really want to write up code to handle that...especially since we would (probably) have to deal with cases where the original certs were generated by cert-manager but the user has indicated that openssl should be used now.  

So, I think the take-aways here are: 

- do nothing if the certs were not generated by us, regardless of their expiration date.
- handle cert-manager and openssl generated certs the same way (and ignore the cert-manager functionality that renews certs for us).

bin/tls-include.sh
```diff
@@ -105,15 +107,22 @@ function create_tls_certs_cm {
   for app in "${apps[@]}"; do
     # Only create the secrets if they do not exist
```
gsmith-sas, line 108: This comment needs to be updated to reflect the new behavior.

bin/tls-include.sh
```diff
@@ -256,8 +265,14 @@ function create_tls_certs_openssl {
       secretName="${app}-tls-secret"
 
       if [ -n "$(kubectl get secret -n $namespace $secretName -o name 2>/dev/null)" ]; then
-         log_debug "TLS Secret for [$app] already exists; skipping TLS certificate generation."
-         continue
+        if kubectl get secret -n $namespace $secretName -o "jsonpath={.data['tls\.crt']}" | base64 -d | openssl x509 -checkend $DAY_IN_SECONDS -noout 2>/dev/null
```
gsmith-sas, line 268: See comments above since they apply here as well.  I wonder if there's value in creating a function to determine whether cert is expired.  That would allow us to call the same function in both places and allow us to update the expiration criteria in a single place.  

In any case, let's use a variable name like CERT_EXPIRATION rather than DAY_IN_SECONDS, since we may, at some point, decide we want the expiration period to be something other than 24 hours.

Hmmmm, if we have different functions for cert-manager vs. openssl, maybe we can implement more sophisticated logic that takes into account cert-manager's ability to regenerate certs.  Let's discuss off-line.

gsmith-sas (changes requested):
Sorry, you were not made aware of some of the wrinkles...see comments and let's discuss offline.

gsmith-sas:
> Couple of things I've considered but want some feedback on:
> 
>     1. Currently will check if cert is expired or will expire within 24 hours. Should the timeframe be different and should that value be configurable by the user(Not sure how much use that would get)?

I think 24 hours is too narrow of a window, maybe something like a week seems more reasonable.  Making it a user-configurable thing might be a good idea even if 99% of users leave the default.  On the other hand, users with special needs could always just delete the existing certs to force us to regenerate.  Let's discuss.

> 
>     2. Should there be a flag that tells the script to not delete and rotate the certs even if they are expired? (This will only happen if TLS_ENABLE=True and the certs already exist)

I am not sure I can identify a reasonable use-case for such a flag.
> 
>     3. Should there be another entrypoint to allow certs to be rotated without having to run the deploy script?

I think you are asking about (suggesting) a stand-alone script that would handled this without actually deploying stuff.  Hmmm, I can see the value in such a script.  On the other hand, we really would like people to regularly update to newer versions of Viya Monitoring.  Would providing such a script make it easier for them to NOT do that?  Hmmm, updating a deployment does NOT currently force an update of certs.  Maybe that is something we should consider as well.  Let's discuss.

bin/renew-tls-certs.sh
```diff
@@ -0,0 +1,178 @@
...
+        done
+
+        log_info "Generating new certs for [Alertmanager, Prometheus, Grafana]"
+        create_tls_certs_openssl "$MON_NS" prometheus alertmanager grafana
+
+        restart-resources "ALL-MON" # WIP: Move restarts to create_tls_certs_openssl function?
+        ;;
+     "ALL-LOG")
+        log_info "Generating new certs for [Alertmanager, Prometheus, Grafana]"
+        log_info "Deleting existing secrets for [Alertmanager, Prometheus, Grafana]"
```
gsmith-sas, line 122: Wrong product names...should be: OpenSearch and OpenSearch Dashboards.

gsmith-sas (changes requested):

gsmith-sas (approved):
Looks good.

PR #453, opened 2023-01-20
[UPGRADE] OpenSearch/OpenSearch Dashboards 2.x

gsmith-sas:


logging/bin/change_internal_password.sh
```diff
@@ -110,9 +112,9 @@ if [[ $response == 4* ]]; then
       ES_ADMIN_PASSWD=$(kubectl -n $LOG_NS get secret internal-user-admin -o=jsonpath="{.data.password}" |base64 --decode)
 
       # make sure hash utility is executable
-      kubectl -n $LOG_NS exec $targetpod --  chmod +x $toolsrootdir/tools/hash.sh
+      kubectl -n $LOG_NS exec $targetpod -c $targetcontainer --  chmod +x $toolsrootdir/tools/hash.sh
       # get hash of new password
-      hashed_passwd=$(kubectl -n $LOG_NS exec $targetpod --  $toolsrootdir/tools/hash.sh -p $NEW_PASSWD)
+      hashed_passwd=$(kubectl -n $LOG_NS exec $targetpod  -c $targetcontainer --  $toolsrootdir/tools/hash.sh -p $NEW_PASSWD|grep -v '*')
       rc=$?
```
gsmith-sas, line 118: In OpenSearch 2.4.1, this utility now outputs a multi-line message indicating that it is deprecated and will be removed in the future prior to outputting the hashed password.  Thankfully, the message lines all start with a "#".

logging/opensearch/bin/run_securityadmin.sh
```diff
@@ -27,3 +27,5 @@ while [ $rc != 0 ] && [ $counter -lt 3 ]; do
   echo "RC: $rc"
   ((counter++))
 done
+
+exit $rc
```
gsmith-sas, line 31: Interestingly, without this "exit" statement, the script would often end with a non-zero exit code (1) that appears to be triggered by the while loop.  It wasn't really causing up problems but the DAC folks have reported an intermittent problem caused by something in our code exiting with a non-zero exit code.  But there was no messages indicating what/where the failure was happening.  Not sure if this will fix it, but it at least ensures our script exits with an rc that accurately reflects the status of the command we executed.

samples/generic-base/logging/user-values-osd.yaml
```diff
@@ -1,12 +1,11 @@
-# Chart: https://github.com/opensearch-project/helm-charts/tree/main/charts/opensearch-dashboards
```
gsmith-sas, line 1: How do you feel about including the version-specific link to the charts/values files?  Part of me would like to keep these links "generic" but given that our code is using a specific version, it seems like we should point them to that version.  On the other hand, this results in another place we need to update Helm chart version numbers...which complicates maintenance.  Thoughts?

cumcke (approved):
Looks good to me.

PR #455, opened 2023-01-25
Replace ghost tunnel proxy

ceelias:


CHANGELOG.md
```diff
@@ -29,7 +30,7 @@
     OpenSearch Dashboards 2.4.1.
   * [FIX] On OpenShift, the deployment order of OpenSearch and OpenSearch Dashboards was reversed to resolve a timing
     issue related to the shared serviceAccount.
-  * [UPGRADE] Introduced new **experimental** getlogs script written in python, to retrieve Opensearch logs with filters and save to a file.
+  * [UPGRADE] Introduced new **experimental** getlogs script written in python, to retrieve OpenSearch logs with filters and save to a file. Relevant documentation can be found [here](./logging/Export_Logs.md#export-logs-python-script---experimental)
```
ceelias, line 33: Added this small fix to the changelog entry for python get logs

samples/tls/path-based-ingress/monitoring/user-values-prom-operator.yaml
```diff
@@ -50,6 +50,14 @@ grafana:
 #   prometheusSpec:
 #     routePrefix: /prometheus
 #     externalUrl: https://host.cluster.example.com/prometheus
+#     alertingEndpoints:
```
ceelias, line 53: Now that I think about it, if a user had `TLS_ENABLE='true'` with `MON_TLS_PATH_INGRESS='true'` and did and update, without updating their configuration, prometheus alerts would not make their way to alertmanager. This section would need to be included. 

Thoughts?

monitoring/tls/prom-alertendpoint-host-https.yaml
```diff
@@ -0,0 +1,9 @@
+prometheus:
+  prometheusSpec:
+    alertingEndpoints:
+      - name: v4m-alertmanager 
+        namespace: monitoring
```
gsmith-sas, line 5: We can't hard-code the namespace here.  Users can deploy our components into any namespace.

monitoring/tls/prom-alertendpoint-path-https.yaml
```diff
@@ -0,0 +1,10 @@
+prometheus:
+  prometheusSpec:
+    alertingEndpoints:
+    - name: v4m-alertmanager 
+      namespace: monitoring
```
gsmith-sas, line 5: See note above re: namespace

gsmith-sas (changes requested):
Looks good EXCEPT this change includes hard-coded references to the "monitoring" namespace.  That's not acceptable since users can (and do) deploy our monitoring components into different namespaces.

gsmith-sas (approved):
LGTM

PR #456, opened 2023-01-26
Automountfix

gsmith-sas:


bin/common.sh
```diff
@@ -187,9 +187,42 @@ function randomPassword {
...
+  sa_name=$2
+  should_disable=${SEC_DISABLE_SA_TOKEN_AUTOMOUNT:-true}
+
+  if [ "$should_disable" == "true" ]; then
+     log_debug "Disabling automount of API tokens for serviceAccount [$ns/$sa_name]"
+     kubectl -n $ns patch serviceAccount $sa_name -p '{"automountServiceAccountToken":false}'
+  fi
+}
+
+function patch_pod_token_automount {
```
gsmith-sas, line 202: Rename function to "enable_pod_token_automount"?  That would make it clear that it is enabling something?

logging/bin/deploy_logging.sh
```diff
@@ -31,8 +31,10 @@ checkDefaultStorageClass
 # Create namespace if it doesn't exist
 if [ "$(kubectl get ns $LOG_NS -o name 2>/dev/null)" == "" ]; then
   kubectl create ns $LOG_NS
-fi
 
+  #Container Security: Disable serviceAccount Token Automounting
+  disable_sa_token_automount $LOG_NS default
```
gsmith-sas, line 36: Currently, only doing it if we're creating the namespace...should we do it always?  Would "patch" existing deployments if doing an UIP.  If user didn't want that to happen, they could set SEC_DISABLE_SA_TOKEN_AUTOMOUNT to "false".  Or, they could manually patch things after the fact.  Thoughts?

bin/common.sh
```diff
@@ -187,9 +187,42 @@ function randomPassword {
...
+function disable_sa_token_automount {
+  local ns sa_name should_disable
+  ns=$1
+  sa_name=$2
+  should_disable=${SEC_DISABLE_SA_TOKEN_AUTOMOUNT:-true}
+
+  if [ "$should_disable" == "true" ]; then
+     log_debug "Disabling automount of API tokens for serviceAccount [$ns/$sa_name]"
+     kubectl -n $ns patch serviceAccount $sa_name -p '{"automountServiceAccountToken":false}'
+  fi
```
gsmith-sas, line 199: Add a DEBUG message indicating that serviceAccount will NOT be patched when SEC_DISABLE_SA_TOKEN_AUTOMOUNT is set to "false"?

bin/common.sh
```diff
@@ -187,9 +187,42 @@ function randomPassword {
...
+
+  if [ "$should_disable" == "true" ]; then
+     log_debug "Enabling automount of API tokens for pods deployed via [$resource_type/$resource_name]"
+
+     if [ "$resource_type" == "daemonset" ] || [ "$resource_type" == "deployment" ]; then
+        kubectl -n $ns patch $resource_type  $resource_name -p '{"spec": {"template": {"spec": {"automountServiceAccountToken":true}}}}'
+     else
+        log_error "Invalid request to function [${FUNCNAME[0]}]; unsupported resource_type [$resource_type]"
+        return 1
+     fi
```
gsmith-sas, line 217: Add a DEBUG message indicating that pod definition will NOT be patched when SEC_DISABLE_SA_TOKEN_AUTOMOUNT is set to "false"?

bin/common.sh
```diff
@@ -187,9 +187,42 @@ function randomPassword {
...
 }
 
+function disable_sa_token_automount {
+  local ns sa_name should_disable
+  ns=$1
+  sa_name=$2
+  should_disable=${SEC_DISABLE_SA_TOKEN_AUTOMOUNT:-true}
+
+  if [ "$should_disable" == "true" ]; then
+     log_debug "Disabling automount of API tokens for serviceAccount [$ns/$sa_name]"
```
gsmith-sas, line 197: Add a check that the ServiceAccount actually exists and exit w/WARNING if not?  Or, let it ERROR out?

bin/common.sh
gsmith-sas: Done

bin/common.sh
gsmith-sas: Done

logging/bin/deploy_logging.sh
cumcke: I would imagine that this would be fine to create if we're creating the namespace.  If the customer's doing an UIP, I could see them wanting to preserve the previous one (unless that could become troublesome).

What's our stance on customer's deploying cluster logging or monitoring in an existing namespace?  Do we frown upon it or is it deemed okay?

If we tell our customers that it's okay to deploy logging in an existing namespace, then this behavior could be problematic.  If not, then this concern can be disregarded.

gsmith-sas: Users are free to deploy Viya Monitoring into any namespace they want.  That said, I think we'd prefer/recommend they keep the Viya Monitoring stuff separated from other applications (e.g. don't deploy Viya Monitoring within a namespace that is being used for SAS Viya).  Some users deploy all of Viya Monitoring (both metric and log monitoring components) into a single namespace.  That shouldn't be a problem either. 

As for UIP scenarios, the only difference between those and a "clean" deploy is that this update will "skip" setting the automount property for the default serviceAccount in that case.  But that won't cause any problems.  We _could_ move the line of code that does that to outside of the IF-block that currently only does it when creating the namespace.  I suspect most UIP scenarios are updating a deployment where we originally created the namespace as opposed to users having done that.  So, in those cases, setting the automount property would probably be just fine.  But just in case the user created the namespace and may have special needs, it seemed better to skip setting that property in those scenarios.

cumcke (approved):
Looks good to me.

PR #458, opened 2023-01-30
[FIX] Update recording rules to capture SAS job metrics with updated naming convention

cumcke:
This change to the query may not work for custom job names.

### DIFF
--- a/monitoring/rules/viya/rules-sas-launcher-job.yaml
+++ b/monitoring/rules/viya/rules-sas-launcher-job.yaml
@@ -9,24 +9,24 @@ spec:
   - name: sas-launcher-job-info.rules
     interval: 6s
     rules:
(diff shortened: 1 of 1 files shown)


gsmith-sas (approved):
I'm surprised that the updated regex doesn't capture too much...looks like it might grab any/every pod that starts with "sas-"...which would seem to include all sorts of non-job pods.  But I'm guessing some of the other criteria limits this to just what we need.  If you're testing hasn't turned any problems, I guess we're "good to go".

cumcke:
> 

I think we're going to be in the clear because the recording rules cross-reference the sas_job_info metric.  If it didn't, I could definitely see everything being flagged as a job in this dashboard.

With that being said, we might even be safe to make it a wildcard (instead of "sas-*") for the same reason, but I'm not familiar enough with the customization option to test it.

PR #468, opened 2023-03-06
[fix] updated kubectl commands to restart pods in change_internal_password script

cumcke:
I also corrected the **Metrics** label under "Unreleased"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
   * [UPGRADE] - K8s-sidecar used with Grafana has been upgraded from 1.19.5 to 1.22.0
 
 * **Logging**
-  * [FIX] Corrected kubectl commands provided in the logging/bin/change_internal_password.sh to reset Fluent Bit and Elastic Exporter pods.
+  * [FIX] Corrected messages in logging/bin/change_internal_password.sh showing commands to restart Fluent Bit and Elastic Exporter pods.
 
 ## Version 1.2.10 (14FEB2023)
 * **Overall**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* **Montoring**
+* **Metrics**
   * [UPGRADE] - Kube-prometheus-stack has been upgraded from version 41.7.3 to 43.3.1
   * [UPGRADE] - Prometheus has been upgraded from version 2.39.0 to 2.40.7
   * [UPGRADE] - Prometheus Operator has been upgraded from version 0.60.0 to 0.62.0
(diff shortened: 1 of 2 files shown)


logging/bin/change_internal_password.sh
```diff
@@ -249,7 +249,7 @@ if [ "$success" == "true" ]; then
           log_notice " the Fluent Bit pods to ensure log collection is not interrupted.            "
           log_notice "                                                                             "
           log_notice " This can be done by submitting the following command:                       "
-          log_notice "         kubectl -n $LOG_NS delete pods -l 'app=fluent-bit, fbout=es'        "
+          log_notice "       kubectl -n logging delete pods -l 'app.kubernetes.io/name=fluent-bit, fbout=es'"
```
gsmith-sas, line 252: - Hard coded namespace name!
- Is there a syntax for filtering using the old and new labels?  I don't recall when the label changed but I suspect it is possible that someone is running an older version (using the app=fluent-bit label) when they run this.  **NEVER MIND...see next bullet**
- Killing the pods to get them to restart was/is a hack.  Let's use the `rollout resetart` command; it's cleaner and not dependent on pod labels.  The command to show is:  `kubectl -n logging rollout restart daemonset v4m-fb`
- (Of course, we don't want to hard-code the namespace name in that command)

logging/bin/change_internal_password.sh
```diff
@@ -278,7 +278,7 @@ if [ "$success" == "true" ]; then
           log_notice " Elasticsearch Exporter pod to ensure OpenSearch metrics continue to be collected. "
           log_notice "                                                                                      "
           log_notice " This can be done by submitting the following command:                                "
-          log_notice "           kubectl -n $LOG_NS delete pod -l 'app=elasticsearch-exporter' "
+          log_notice "       kubectl -n logging delete pods -l 'app=prometheus-elasticsearch-exporter'      "
```
gsmith-sas, line 281: - Hard-coded namespace name
- Use rollout restart here too?  In this case, the pod is deployed as part of a deployment (named v4m-es-exporter) rather than as part a daemonset, so the syntax is slightly different.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
   * [UPGRADE] - Pushgateway has been upgraded from version 1.4.3 to 1.5.1
   * [UPGRADE] - K8s-sidecar used with Grafana has been upgraded from 1.19.5 to 1.22.0
 
+* **Logging**
+  * [FIX] Corrected kubectl commands provided in the logging/bin/change_internal_password.sh to reset Fluent Bit and Elastic Exporter pods.
```
gsmith-sas, line 17: Slight tweak to clarify what was changed and correct the component name: 
[FIX] Corrected messages in logging/bin/change_internal_password.sh showing commands to restart Fluent Bit and Elasticsearch Exporter pods.

gsmith-sas (changes requested):
A tweak of the CHANGELOG message...and I realized we could use a cleaner kubectl command to restart the pods.

ceelias (approved):

gsmith-sas (approved):

PR #476, opened 2023-03-20
[Upgrade] OpenSearch/OpenSearch Dashboards 2.6.0; Fluent Bit 2.0.9

gsmith-sas:


logging/bin/deploy_osd_content.sh
```diff
@@ -65,7 +65,8 @@ set +e  # disable exit on error
 # Need to wait 2-3 minutes for OSD to come up and
 # and be ready to accept the curl commands below
 # Confirm OSD is ready
-for pause in 30 30 60 30 30 30 30 30 30
+log_info "Waiting (up to more 8 minutes) for OpenSearch Dashboards API endpoint to be ready"
+for pause in 30 30 60 30 30 30 30 30 30 60 60 60
```
gsmith-sas, line 69: I ran into a problem where, on occasion, OpenSearch pods were still restarting after the upgrade when this script ran.  There wasn't a more elegant way to test for that, so I added another couple pause loop executions to the existing pause here.  In most cases, I don't think it will have any impact (since things will be ready) but I think it will handle the slowness I encountered on some occasions on some platforms.

ceelias (approved):

ceelias (approved):

PR #482, opened 2023-04-10
Getlogs update

vishaanS:
COMPUTESVCS - 48010

### DIFF
--- a/logging/bin/getlogs.py
+++ b/logging/bin/getlogs.py
@@ -292,7 +292,7 @@ if (args['out-filename']): ##Check if user specified file exists, and if it shou
         print("Search complete. Results printed to " + args['out-filename'])    
 else:
     print("Search complete.")
-    if (args['format'].equals("json")):
+    if args['format'] == "json":
         print(json.dumps(response['hits']['hits'], sort_keys=True, indent=2))
     else:
         print(str(response['hits']['hits']))
--- a/logging/bin/getlogs.py
+++ b/logging/bin/getlogs.py
@@ -15,10 +15,10 @@ if os.path.isfile("temp_SAS_OSQ.TXT"):
 
 def validate_input(dict):
     """Validate the arguments passed by the user to ensure script will function properly"""
-    if args['create-auth']:
+    if dict['create-auth']:
         if (type(dict['create-auth'] == list)):
             dict['create-auth'] = " ".join(dict['create-auth'])
-        x = open(args['create-auth'], "w")
+        x = open(dict['create-auth'], "w")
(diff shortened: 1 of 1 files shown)


logging/bin/getlogs.py
```diff
@@ -15,29 +15,34 @@
 
 def validate_input(dict):
     """Validate the arguments passed by the user to ensure script will function properly"""
-    if args['create-auth']:
+    if dict['create-auth']:
```
ceelias, line 18: This change made me think a little bit about some of the var names we are using in this script. Lets rename this list to something more descriptive. (maybe something like `arg_list` or whatever you feel can convey what its being used for)

logging/bin/getlogs.py
```diff
@@ -15,29 +15,34 @@
...
+            dict['port'] = authDict['port']
         except Exception as e:
             print("There was a problem with the authentication file. Please ensure that the values are in dictionary format with keys: 'username', 'password', 'host', 'port'")
             print(e)
             exit()
 
-    
+    """Ensure maxrows is less than 10000"""
+
+    if dict['maxInt'] > 10000:
```
ceelias, line 42: Although its fairly clear whats happening here, lets get rid of the magic number here and define a constant. Could be useful if in the future we want to make it configurable but really just helps with maintainability/readability with the code 

```python
MAX_ROWS = 1000

if dict['maxInt'] > MAX_ROWS:
```

logging/bin/getlogs.py
```diff
@@ -15,29 +15,34 @@
...
+        print("Error: Maxrows limit of 10000 exceeded.")
```
ceelias, line 43: Might be worth creating an error function that takes in a string and then exits. That could help standardize the error messages and make it easier to find all the places we are throwing an error. I notice there is an error thrown on line 36 which doesn't use  the `Error: ` prefix you used here. This could also remove some duplicate code by having to call print (x) and exit() everywhere you want to throw

logging/bin/getlogs.py
```diff
@@ -47,39 +52,49 @@ def validate_input(dict):
     if(len(sys.argv) == 1): ##Check if any args have been provided
```
ceelias, line 52: Doesn't need to happen right now but I think when I moved this code over from gitlab and updated some things that we no longer need this check, I'm pretty sure args are already grabbed by the time we've made it this far in the code. Def could be wrong here though and this isn't something that needs to happen now, but we should likely make note of this and clean it up if so.

ceelias (commented):
Looks good! I commented on some things that might be worth updating at some point. None are critical and can be addressed at a later time if you would prefer.

PR #483, opened 2023-04-12
Added check to ensure user workload monitoring is enabled for openshift

ceelias:


monitoring/bin/deploy_monitoring_viya.sh
```diff
@@ -7,6 +7,17 @@ cd "$(dirname $BASH_SOURCE)/../.."
 source monitoring/bin/common.sh
 source bin/openshift-include.sh
 
+# If openshift deployment, ensure user-workload monitoring is enabled
+if [ -n "$(kubectl get ns openshift-monitoring -o name 2>/dev/null)" ]; then
```
ceelias, line 11: chose `openshift-monitoring` namespace as that should only exist in the event of an openshift deployment with monitoring deployed. 

Any concerns about choosing this one?

gsmith-sas: Is it even possible to deploy OpenShift without the built-in Prometheus?  Our tooling assumes that the built-in Prometheus is "there".  If it is possible that is not the case, we probably should be handling that scenario (but not in this current PR/effort).

monitoring/bin/deploy_monitoring_viya.sh
```diff
@@ -7,6 +7,17 @@ cd "$(dirname $BASH_SOURCE)/../.."
...
+  if [ -n "$(kubectl get po -n openshift-user-workload-monitoring -o name 2>/dev/null)" ]; then
```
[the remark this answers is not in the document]
gsmith-sas: Oh, so the existence of the namespace is **not** a sign that workload monitoring has been enabled....good catch!

monitoring/bin/deploy_monitoring_viya.sh
```diff
@@ -7,6 +7,17 @@ cd "$(dirname $BASH_SOURCE)/../.."
...
+    log_debug "User workload monitoring is enabled"
+  else
+    log_error "Ensure that user workload monitoring is enabled"
+    log_error "openshift-user-workload-monitoring namespace was not found"
+    exit 1
```
gsmith-sas, line 17: I wonder if "ERROR and exit" is the right behavior.  As it stands now, fixing the problem (i.e. updating the appropriate configMap) after running our script will fix the problem: there is no need to re-run the deploy_monitoring_viya.sh script.  Given that, maybe a WARNING is more appropriate? Although "ERROR and exit" should mean they can't miss the message.  So I can argue either side.  I also wonder if we should provide more information on how to "fix" the problem (e.g. tell them to do x-y-z, or, point them to the documentation)?

gsmith-sas (commented):
Looks good but see my comments and let's discuss.

PR #488, opened 2023-04-20
[fix] addressed issue Prometheus data source on path-based ingress TLS deployments

cumcke:


monitoring/tls/prom-alertendpoint-path-https.yaml
```diff
@@ -1,9 +1,10 @@
...
   prometheusSpec:
     alertingEndpoints:
-      - name: v4m-alertmanager 
+      - name: v4m-alertmanager
         port: http-web
         pathPrefix: "/alertmanager"
         scheme: https
         tlsConfig:
           insecureSkipVerify: true
+    routePrefix: /prometheus
```
ceelias, line 10: This was previously being defined in the sample right? No problem with this change, just curious about the motive

cumcke: Yep!  This was originally going to be a change we made to the samples, but it made them ... a bit more confusing to use.  It would mean that, even if they did not want to expose a URL for Prometheus, they'd still have to uncomment 3 or so lines in the samples (instead of an all or nothing situation).
With this change, we can preserve the all-or-nothing approach to the samples such that, if they uncomment or change the routePrefix in the sample, the USER_DIR one would still take priority.

ceelias: Cool sounds good! Does it make sense to just remove it from the sample then? I think both TLS and Ingress samples have it

ceelias (approved):

PR #490, opened 2023-04-28
GetLogs file permissions work, standardizing output directory.

vishaanS:


logging/bin/getlogs.py
```diff
@@ -9,22 +9,34 @@
 import os, sys
 import json, csv
 
-## Cleanup temp file if not previously deleted
-if os.path.isfile("temp_SAS_OSQ.TXT"):
-    os.remove("temp_SAS_OSQ.TXT")
+## Cleanup temp file if not previously deleted, Create target directory
+if not os.path.isdir("GetLogs_Files"):
```
ceelias, line 13: In the future lets make an update to make these file names a variable, so we will only have to update them in one place if need be and will avoid mispellings that can cause issues.

logging/bin/getlogs.py
```diff
@@ -9,22 +9,34 @@
...
     """Validate the arguments passed by the user to ensure script will function properly"""
+    
+    if(len(sys.argv) == 1): ##Check if any args have been provided
+        print("No arguments have been provided")
+        exit()
+
     if dict['create-auth']:
         if (type(dict['create-auth'] == list)):
             dict['create-auth'] = " ".join(dict['create-auth'])
+        if (not "GetLogs_Files/" in  dict['create-auth']):
```
ceelias, line 29: A little confused about the need for this check. Is the user expected to include the directory name with the file name? If so, what will happen if they are passing in a different directory name? Maybe its fine but just wanted to pose the question.

logging/bin/getlogs.py
```diff
@@ -92,7 +107,21 @@ def validate_input(dict):
         if ((not ".json" in dict['savequery'])):
             print('Error: Not a supported filetype for the saved query file.')
             exit()
+        if (not "GetLogs_Files/" in dict['savequery']):
+            dict['savequery']= os.path.join("GetLogs_Files", dict['savequery'])
 
+    if dict['query-filename']:
+        if (not "GetLogs_Files/" in dict['query-filename']):
```
ceelias, line 114: Again, future update, but since we keep doing this check for directory name and update if the directory isn't present maybe we should turn this into a function.

ceelias (commented):
Looks good otherwise. Most things I poiinted out can be updates at a later time to improve quality of life.

PR #491, opened 2023-05-08
standardized viya naming

technomerlyn:


CHANGELOG.md
```diff
@@ -131,7 +131,7 @@
   * [FIX] Updated the Dockerfile so that the subdirectories are preserved in the Docker container.
 
 * **Metrics**
-  * [FIX] Added Service Monitor definition to permit metric collection from OpenSearch instance within individual SAS Viya deployments.
+  * [FIX] Added Service Monitor definition to permit metric collection from OpenSearch instance within individual SAS Viya platform deployments.
```
gsmith-sas, line 134: The service monitor definitions will permit metric collection from any OpenSearch instance deployed within any flavor of SAS Viya, it is not limited to those instances deployed with SAS Viya platform.  Given that, the addition of "platform" seems inappropriate.

CHANGELOG.md
```diff
@@ -203,7 +203,7 @@
   * [CHANGE] - Running scripts on OpenShift 4.9 no longer generates a WARNING message.
 
 * **Monitoring**
-  * [FIX] - Updated SAS Viya logo on the Grafana Welcome screen so that it is easier to see on darker background.
+  * [FIX] - Updated SAS Viya platform logo on the Grafana Welcome screen so that it is easier to see on darker background.
```
gsmith-sas, line 206: Is the logo the "SAS Viya platform" logo or the "SAS Viya logo"?  I suspect/assume it is the latter.  If so, the addition of "platform" would be inappropriate.

CHANGELOG.md
```diff
@@ -277,7 +277,7 @@
 ## Version 1.1.5 (15FEB22)
 
 * **Monitoring**
-  * [CHANGE] - The SAS Viya Welcome dashboard has been updated with a
+  * [CHANGE] - The SAS Viya platform Welcome dashboard has been updated with a
```
gsmith-sas, line 280: The name of this dashboard (as shown in the list of Grafana dashboards and on the dashboard itself) is "SAS Viya Welcome".  If we need to revisit/revise that, we can discuss.  But, unless it changes, I think the reference here should match what the user will see in Grafana.

CHANGELOG.md
```diff
@@ -613,7 +613,7 @@
   supported to supply user-provided dashboards at deployment time
   * [FEATURE] The new [CloudWatch sample](samples/cloudwatch) provides
   instructions on configuring the CloudWatch agent to scrape metrics
-  from SAS Viya components
+  from SAS Viya platform components
```
gsmith-sas, line 616: See general comments.

CHANGELOG.md
```diff
@@ -633,7 +633,7 @@
     * flant-statusmap-panel
     * btplc-status-dot-panel
   * [CHANGE] [cert-manager](https://cert-manager.io/docs/usage/certificate/) resources
-  now use 'v1' to align with their use in SAS Viya 4.x
+  now use 'v1' to align with their use in SAS Viya platform 4.
```
gsmith-sas, line 636: See general comments.  But, in any case, I suspect "SAS Viya platform 4" is never correct.

gsmith-sas, line 636: See general comments.  But, in any case, I suspect "SAS Viya platform 4" is never correct.

samples/README.md
```diff
@@ -49,24 +49,24 @@ These samples are provided:
 * [azure-deployment](azure-deployment) - Deploys on Microsoft Azure Kubernetes
   Service (AKS).
 * [azure-monitor](azure-monitor) - Enables Azure Monitor to collect metrics
```
gsmith-sas, line 51: The SAS Viya Monitoring project deploys components intended to monitor SAS Viya (in all flavors) and not just the SAS Viya platform.  These edits suggest the monitoring is limited to the SAS Viya platform.

gsmith-sas, line 51: The SAS Viya Monitoring project deploys components intended to monitor SAS Viya (in all flavors) and not just the SAS Viya platform.  These edits suggest the monitoring is limited to the SAS Viya platform.

samples/azure-monitor/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Scrape Prometheus Metrics Endpoints
 
-SAS Viya components are natively instrumented to expose a Prometheus-compatible
+SAS Viya platform components are natively instrumented to expose a Prometheus-compatible
```
gsmith-sas, line 5: There is a requirement for all components within SAS Viya (all flavors) to be instrumented to permit monitoring in this way.  Therefore, I believe the addition of "platform" here is inappropriate.

samples/azure-monitor/README.md
```diff
@@ -23,7 +23,7 @@ guide. Recommended changes include:
 - `monitor_kubernetes_pods` - Set to `true`.
 - `monitor_kubernetes_pods` - Set to `true` to enable Azure Monitor to
 auto-discover pods to monitor, based on the standard Prometheus annotations.
-SAS Viya components that expose metrics endpoints should include these
+SAS Viya platform components that expose metrics endpoints should include these
 annotations:
```
gsmith-sas, line 27: see previous comment.

samples/cloudwatch/README.md
```diff
@@ -1,11 +1,11 @@
 # Amazon CloudWatch Integration
 
 This sample describes how to configure Amazon CloudWatch to collect metrics
-from the services and components in SAS Viya. Complete the deployment
+from the services and components in SAS Viya platform platform. Complete the deployment
```
gsmith-sas, line 4: "platform platform" is almost certainly incorrect.  However, see comment above re: scope of instrumentation and monitoring.

samples/gke-monitoring/README.md
```diff
@@ -39,7 +39,7 @@ volume mount so that the sidecar can directly read the Prometheus server data.
 Although there are several ways to deploy the sidecar to a Prometheus pod,
 this sample directly supports only
 [Prometheus custom resources](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheus)
-such as those deployed by the SAS Viya monitoring components in
+such as those deployed by the SAS Viya platform monitoring components in
```
gsmith-sas, line 42: This is a reference to the SAS Viya Monitoring project and not to SAS Viya (itself); therefore addition of "platform" seems inappropriate.

samples/namespace-monitoring/README.md
```diff
@@ -1,18 +1,18 @@
 # Namespace Monitoring
 
 This sample demonstrates how to customize a monitoring
-deployment to separate cluster monitoring from SAS Viya (namespace)
+deployment to separate cluster monitoring from SAS Viya platform (namespace)
 monitoring.
```
gsmith-sas, line 5: This relevance of this sample isn't limited to deployments of "SAS Viya platform"  only; the same approach could be used to handle a similar need for any form of SAS Viya.

gsmith-sas (commented):
@technomerlyn I'm sorry but I think this set of changes will require more work. I've made specific comments next to specific changes but I think most of those fall under a number of general (high-level) concerns, listed below:

- I am not completely comfortable with making changes to the CHANGELOG.md file since it is intended to be a on-going record of changes introduced with each release.  Making significant revisions (other than corrections and/or revising deadlinks) feels "wrong".
- Separate from the above, I believe CHANGELOG entries should use the terminology and product names that were applicable when the dated entries were added. I don't recall when the new product naming took effect, but the CHANGELOG goes back to the start of the project which absolutely pre-dates the product name change.
- The SAS Viya Monitoring project deploys components to monitor SAS Viya, in all of its forms.  It is not limited to nor intended to only monitor the SAS Viya platform.  Therefore, changing every reference of "SAS Viya" to "SAS Viya platform" seems both inappropriate and, more importantly, inaccurate.
- The previous point is especially relevant to the various samples included in our project.  Adding "platform" to the various references to "SAS Viya" could lead users to assume that the sample is not relevant to their use case because they are using some other flavor of SAS Viya.
- As explained in a specific comment below, the instrumentation to permit metric and log monitoring was/is not a requirement limited to SAS Viya platform components; the scope of that requirement is intended to apply to all components of all flavors of SAS Viya 4.

Take a look at the comments and then let's talk.

PR #500, opened 2023-05-18
Update Kubernetes/OpenShift version checking to reflect current info

gsmith-sas:


bin/openshift-include.sh
```diff
@@ -64,12 +64,12 @@ function ocVersionCheck {
     ### Openshift versions that do not start with a 4 should produce an error.
     if (( "$OC_MAJOR_VERSION" != 4 )); then
       log_error "Unsupported OpenShift client version: $OC_FULL_VERSION"
-      log_error "Version 4.8+ is required"
+      log_error "Version 4.9+ is required"
```
ceelias, line 67: No real issue just think its interesting that the error msg is the same as the one below, so maybe at somepoint we should just combine the version checks into a single IF block.

ceelias (approved):

PR #505, opened 2023-06-15
Uplogging

gsmith-sas:


logging/user-values-opensearch.yaml
```diff
@@ -1,8 +1,8 @@
-# Chart: https://github.com/opensearch-project/helm-charts/tree/main/charts/opensearch
```
gsmith-sas, line 1: Evidently, this file was not updated when we did our last few updates; that's why the diff shows so many changes.

logging/bin/change_internal_password.sh
```diff
@@ -155,7 +155,8 @@ if [[ $response == 4* ]]; then
       # make sure hash utility is executable
       kubectl -n $LOG_NS exec $targetpod  -c $targetcontainer --  chmod +x $toolsrootdir/tools/hash.sh
       # get hash of new password
-      hashed_passwd=$(kubectl -n $LOG_NS exec $targetpod  -c $targetcontainer --  $toolsrootdir/tools/hash.sh -p $NEW_PASSWD)
```
gsmith-sas, line 158: Several months back, we had to add the grep command to a similar/identical command (line 117) b/c OpenSearch decided to add a deprecation notice in the output of the hash.sh script.  We should have done it here as well and didn't.  Correcting that oversight now.

logging/opensearch/securityconfig/config.yml
```diff
@@ -65,9 +65,11 @@ config:
     # Set filtered_alias_mode to 'nowarn' to allow more than 2 filtered aliases per index silently
     #filtered_alias_mode: warn
     do_not_fail_on_forbidden: true
-    #kibana:
-    # Kibana multitenancy
-    #multitenancy_enabled: true
+    kibana:
+    # Kibana multitenancy 
+      multitenancy_enabled: true
```
ceelias, line 70: I know this section was enabled to disable private tenant but does multitenancy need to be enabled as well?

samples/ingress/host-based-ingress/user.env
```diff
@@ -0,0 +1,9 @@
+# This sample demonstrates how to configure
```
ceelias, line 1: These seem a bit excessive to show how to set this single variable but I could be wrong. I think it would be helpful to create a user.env file that shows all the variables and the defaults so that it is easy to see what is configurable similar to how the helm configuration files are done but we can discuss that at a later time.

ceelias (approved):

logging/opensearch/securityconfig/config.yml
gsmith-sas: This flag relates to an OpenSearch Dashboards (OSD) concept of tenancy and NOT to the SAS Viya concept of tenancy.  So, we are disabling a feature in OSD that gave every user a private area for their own content while also enabling the overall OSD features that allow tenant-specific content to be isolated from other tenants.  The OSD 'private' tenant space could be a nice feature but given its current implementation in OSD, it more likely to lead to confusion rather than be helpful.

samples/ingress/host-based-ingress/user.env
gsmith-sas: Agreed.  The TLS and the Ingress samples need to be re-worked given our shift to enabling TLS by default.  I'm working on that and will (probably) submit changes in a separate PR.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -270,9 +270,9 @@ set +e
 get_ingress_ports
 
 # get URLs for Grafana, Prometheus and AlertManager
-gf_url=$(get_service_url $MON_NS v4m-grafana  "false")
-# pr_url=$(get_url $MON_NS v4m-prometheus  "false")
-# am_url=$(get_url $MON_NS v4m-alertmanager  "false")
+gf_url=$(get_service_url $MON_NS v4m-grafana  "$TLS_ENABLE")
+# pr_url=$(get_url $MON_NS v4m-prometheus  "$TLS_ENABLE")
```
ceelias, line 274: Just out of curiosity, should these be commented out, or is there a plan to uncomment in the future?

gsmith-sas: We used to make these other web applications accessible by default.  However, we stopped doing that a while back since they lack any built-in authentication capability.  I suspect we should probably just remove the commented out lines but didn't feel comfortable doing that without further discussion/thought.

PR #509, opened 2023-06-28
Changes to sas-opendistro ServiceMonitor so it can be used with prometheus exporter plugin for OpenSearch

dancristiancecoi:
Modified the ServiceMonitor definition so it uses basicAuth to connect to the exporter endpoint & added relabelling so it gets the scheme, path etc from Opendistro's annotations

### DIFF
--- a/monitoring/monitors/viya/serviceMonitor-sas-opensearch.yaml
+++ b/monitoring/monitors/viya/serviceMonitor-sas-opensearch.yaml
@@ -39,4 +39,4 @@ spec:
       insecureSkipVerify: true
   selector:
     matchLabels:
-      app.kubernetes.io/name: sas-opendistro
+      opendistro.sas.com/service-name: sas-opendistro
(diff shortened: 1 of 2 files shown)


monitoring/monitors/viya/serviceMonitor-sas-opensearch.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: sas-opensearch
+spec:
+  endpoints:
+  - basicAuth:
+      password:
+        key: password
+        name: sas-opendistro-sasadmin-secret
+      username:
+        key: username
+        name: sas-opendistro-sasadmin-secret
```
gsmith-sas, line 13: Adding **port: https** here (after line13, indentation matching metricRelabelings and basicAuth) appears to limit the Prometheus scraping to only the desired port.

gsmith-sas (approved):
Looks good to me.  Thanks for the contribution!

PR #510, opened 2023-06-29
Improved log message handling for some PostgreSQL variants

gsmith-sas:


logging/fb/fluent-bit_config.configmap_azmonitor.yaml
```diff
@@ -4,7 +4,7 @@ data:
     [FILTER]
        Name modify
        Match *
-       Set  fb_configMap_version 0.2.23
```
ceelias, line 7: Whats the meaning of the `k` in the version here?

gsmith-sas: It is an revision identifier...the first iteration of changes during this effort were identified as "0.2.24a", the next "0.2.24b", etc.  During testing, we need this field to validate which configuration was used to process the log message.  We have an end of sprint process that will update this version number to the correct value (which we expect to be "0.2.24").

ceelias (approved):

PR #513, opened 2023-07-11
[feature] Add support for air-gapped environments

cumcke:


logging/bin/deploy_opensearch.sh
```diff
@@ -328,6 +358,7 @@ if [ "$deploy_temp_masters" == "true" ]; then
        --namespace $LOG_NS \
        --values logging/opensearch/opensearch_helm_values.yaml \
        --values "$wnpValuesFile" \
+       --values "$airgapValuesFile" \
```
gsmith-sas, line 361: Don't we need/want to reference $AIRGAP_HELM_REPO below (on line 370)?   Might be good to split line 370 below so the Helm chart name sits on its own line.

logging/bin/deploy_osd.sh
```diff
@@ -143,6 +170,7 @@ helm $helmDebug upgrade --install v4m-osd \
     --values logging/opensearch/osd_helm_values.yaml \
     --values "$wnpValuesFile" \
     --values "$nodeport_yaml" \
+    --values "$airgapValuesFile" \
```
gsmith-sas, line 173: Don't we need/want to reference $AIRGAP_HELM_REPO below (on line 177)?   Might be good to split line 177 below so the Helm chart name sits on its own line.

logging/bin/deploy_fluentbit_opensearch.sh
```diff
@@ -150,6 +177,7 @@ helm $helmDebug upgrade --install --namespace $LOG_NS v4m-fb  \
   --version $FLUENTBIT_HELM_CHART_VERSION \
   --values logging/fb/fluent-bit_helm_values_opensearch.yaml  \
   --values $openshiftValuesFile \
+  --values $airgapValuesFile \
```
gsmith-sas, line 180: Don't we need/want to reference $AIRGAP_HELM_REPO below (on line 182)?   Might be good to split line 182 below so the Helm chart name sits on its own line.

logging/bin/deploy_fluentbit_azmonitor.sh
```diff
@@ -140,6 +168,7 @@ FLUENTBIT_HELM_CHART_VERSION=${FLUENTBIT_HELM_CHART_VERSION:-"0.24.0"}
 helm $helmDebug upgrade --install v4m-fbaz  --namespace $LOG_NS  \
   --version $FLUENTBIT_HELM_CHART_VERSION \
   --values logging/fb/fluent-bit_helm_values_azmonitor.yaml \
+  --values $airgapValuesFile \
```
gsmith-sas, line 171: Don't we need/want to reference $AIRGAP_HELM_REPO below (on line 173)?   Might be good to split line 173 below so the Helm chart name sits on its own line.

logging/bin/deploy_esexporter.sh
```diff
@@ -107,6 +135,7 @@ helm $helmDebug upgrade --install es-exporter \
  -f $primaryValuesFile \
  -f $wnpValuesFile \
  -f $openshiftValuesFile \
+ -f $airgapValuesFile \
```
gsmith-sas, line 138: Don't we need/want to reference $AIRGAP_HELM_REPO below (on line 140)?   Hmmm, not sure why the \-\-set fullnameOverride option appears _after_ the name of the chart.  That's inconsistent with the ordering of args in all of our other Helm chart deployment calls.  Probably should move it to be consistent with the others.

bin/common.sh
```diff
@@ -92,6 +92,36 @@ if [ "$SAS_COMMON_SOURCED" = "" ]; then
        fi
     fi
 
+    ## Check for air gap deployment
+    AIRGAP_DEPLOYMENT=${AIRGAP_DEPLOYMENT:-false}
```
gsmith-sas, line 96: The common.sh file is sourced by just about every script, even those not deploying things.  It might make sense to move the airgap-specific logic to its own include file and only source it into the deploy_*.sh files.   Give that some thought.

gsmith-sas (changes requested):
This looks good.  But we need to make some changes:

- I think the block of code that takes a yaml file and then does the text substitutions can be turned into a function that can be called everywhere it is needed rather than have it repeated several times.  It could take a single argument that identifies the file to be updated and then "transorms" it.  OR, perhaps, have it return either the name of the transformed file or $TMP_DIR/empty_dir.
- It looks like the logic to use the AIRGAP_HELM_REPO env var in the helm install command for the log monitoring components was left out.  So, these changes support pulling the container images from the ACR but not the Helm charts.
- The common.sh file is sourced by just about every script, even those not deploying things.  It might make sense to move the airgap-specific logic to its own include file and only source it into the deploy_*.sh files. 
- A couple of minor tweaks to some message text/formatting
- miscellaneous other things

bin/common.sh
```diff
@@ -92,6 +92,36 @@ if [ "$SAS_COMMON_SOURCED" = "" ]; then
...
+          log_debug "Separate AIRGAP_HELM_REPO value not provided"
+          log_debug "Setting AIRGAP_HELM_REPO to oci://${AIRGAP_REGISTRY}/"
+          AIRGAP_HELM_REPO="oci://${AIRGAP_REGISTRY}/"
+        fi  
+      fi
+    fi
+
+    ## The user will need to create the namespace and secret before running the deployment scripts.
+    ## This function will produce an error if the secret is not found in the environment.
+    function checkForAirgapSecretToNamespace {
```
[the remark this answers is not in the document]
cumcke: The function was supposed to be called checkForAirgapSecretInNamespace ... I think I was in between naming ideas and didn't fully change it.  It will be updated in the next push.

I've also made sure to only call it after the namespace has been made so it has to exist.
Cluster Metric Monitoring - It's positioned after the kubectl create $MON_NS line.
Viya Metric Monitoring - The function's only called after the check to make sure $VIYA_NS exists happens so it will have to exist.
Log Monitoring - Assuming the users are exclusively using deploy_logging.sh, it will be after the kubectl create $LOG_NS.  The only way I can think of a namespace not existing is if the customer tries to run one of the independent scripts on the logging side prior to running the deploy_logging.sh (to create the namespace); in that case, it would throw a kubectl error saying the namespace doesn't exist ... but also confirms that the secret wouldn't be in the namespace because it doesn't exist.  I think that's okay.

bin/common.sh
```diff
@@ -92,6 +92,36 @@ if [ "$SAS_COMMON_SOURCED" = "" ]; then
...
+      secretName="$1"
+      namespace="$2" 
+      if [ -z "$(kubectl get secret -n $namespace | grep $secretName)" ]; then
```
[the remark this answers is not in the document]
cumcke: At the stage where it is called, the namespace will always exist unless they run an independent logging script before running the deploy_logging.sh script.  See earlier comment for more details.

bin/common.sh
```diff
@@ -92,6 +92,36 @@ if [ "$SAS_COMMON_SOURCED" = "" ]; then
...
+        log_error "The secret, $secretName, was not detected"
+        log_error "Please add the Image Pull secret to the $namespace namespace and run the deployment script again"
```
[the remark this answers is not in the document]
cumcke: I updated the message to use "image pull secret" in both places.  It will be updated in the next push.

logging/bin/deploy_osd.sh
cumcke: We do need to add the $AIRGAP_HELM_REPO ... I might've missed this one because it was kind of flushed in like that.  I moved it to its own line and added the $AIRGAP_HELM_REPO parameter to it.  It will show up in the next push.

logging/bin/deploy_fluentbit_azmonitor.sh
cumcke: We do need to add the $AIRGAP_HELM_REPO ... I might've missed this one because it was kind of flushed in like that.  I moved it to its own line and added the $AIRGAP_HELM_REPO parameter to it.  It will show up in the next push.

logging/bin/deploy_esexporter.sh
cumcke: We do need to add the $AIRGAP_HELM_REPO ... I might've missed this one because it was kind of flushed in like that.  I moved it to its own line and added the $AIRGAP_HELM_REPO parameter to it.  It will show up in the next push.

bin/common.sh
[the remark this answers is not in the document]
gsmith-sas: Ok, sounds like you have the namespace exists checking sorted out.  We don't have a formal standard for checking the existence of a specific Kubernetes resource, but we probably should standardize on something like the following (which is used in a number of other places in our scripts):

if kubectl -n $namespace get secret $secretName 2>/dev/null 1>&2; then...

That syntax eliminates the need to pipe multiple commands and handles the unexpected failure of the command better.

logging/bin/deploy_esexporter.sh
```diff
@@ -58,20 +58,48 @@ fi
...
+## Check for air gap deployment
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  
+  # Check for the image pull secret for the air gap environment
+  checkForAirgapSecretToNamespace "$AIRGAP_IMAGE_PULL_SECRET_NAME" "$LOG_NS"
+
+  # Copy template files to temp
+  airgapDir="$TMP_DIR/airgap"
+  mkdir -p $airgapDir
+  cp -R logging/airgap/airgap-values-es-exporter.yaml $airgapDir/
```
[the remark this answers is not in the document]
cumcke: I think it was just a copy/paste job I didn't handle very well.
I've gone ahead and removed the -R and moved that logic to its own function like you suggested.

bin/airgap-include.sh
```diff
@@ -0,0 +1,56 @@
+# Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# This file is not marked as executable as it is intended to be sourced
+# Current directory must be the root directory of the repo
+
+## Check for AIRGAP_REGISTRY, if null/empty, error out.  Otherwise set and create HELM_URL_BASE.
```
gsmith-sas, line 7: Take a look at [bin/common.sh|https://github.com/sassoftware/viya4-monitoring-kubernetes/blob/e542b813ea8c1d5126acd7b69401d84b2d3f1ffa/bin/common.sh#L35] and [bin/openshift-include.sh|https://github.com/sassoftware/viya4-monitoring-kubernetes/blob/main/bin/openshift-include.sh#L81] to see how we ensure we only execute the common code once; we should do something similar here.

logging/airgap/airgap-fluent-bit.yaml
```diff
@@ -0,0 +1,5 @@
+image:
```
gsmith-sas, line 1: I don't think we need to "airgap-" in the names of these files since they are in the "airgap" directory.

logging/bin/deploy_opensearch.sh
```diff
@@ -40,6 +40,20 @@ fi
 # Get/Set Helm Chart Version
 OPENSEARCH_HELM_CHART_VERSION=${OPENSEARCH_HELM_CHART_VERSION:-"2.11.0"}
 
+## Check for air gap deployment
+## Check for air gap deployment
```
gsmith-sas, line 44: Repeated/redundant comment?

logging/bin/deploy_opensearch.sh
```diff
@@ -133,8 +147,10 @@ if [ "$HELM_DEBUG" == "true" ]; then
 fi
 
 helmRepoAdd opensearch  https://opensearch-project.github.io/helm-charts
-log_verbose "Updating Helm repositories"
-helm repo update
+
+## Commenting out as it might be redundant code
+# log_verbose "Updating Helm repositories"
+# helm repo update
```
gsmith-sas, line 154: Remove?

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -73,11 +108,26 @@ if [ "$PROM_OPERATOR_CRD_UPDATE" == "true" ]; then
...
+    ## Determine CRD URL - if in an airgap environment, look for them in USER_DIR.
+    if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+      crdURL=$USER_DIR/monitoring/prometheus-operator-crd/$PROM_OPERATOR_CRD_VERSION/monitoring.coreos.com_$crd.yaml
+
+      ## Fail if the CRDs could not be located.
+      if [ ! -f "$crdURL" ]; then
+        log_error "Unable to locate file: monitoring.coreos.com_$crd.yaml in"
+        log_error "$USER_DIR/monitoring/prometheus-operator-crd/$PROM_OPERATOR_CRD_VERSION directory"
+        log_error "Please make sure to provide all Prometheus Operator CRDs before running the deployment"
+        exit 1
```
gsmith-sas, line 121: I wonder if we should move this validation that the CRD files are in the USER_DIR to the common.sh or airgap-include.sh.  Basically, check early so we can fail early and not fail after partially deploying stuff.  Seems like a better user experience.

monitoring/bin/deploy_monitoring_viya.sh
```diff
@@ -60,7 +60,35 @@ set -e
...
+  ## Check for air gap deployment
+  if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  
+    # Check for the image pull secret for the air gap environment
+    checkForAirgapSecretInNamespace "$AIRGAP_IMAGE_PULL_SECRET_NAME" "$VIYA_NS"
+
+    # Copy template files to temp
+    airgapDir="$TMP_DIR/airgap/viya"
+    mkdir -p $airgapDir
+    cp -R monitoring/airgap/viya/airgap-values-pushgateway.yaml $airgapDir/airgap-values-pushgateway.yaml
```
gsmith-sas, line 74: see note above re: including "airgap" in file names.

logging/bin/deploy_osd.sh
```diff
@@ -29,6 +29,22 @@ set -e
 #Fail if not using OpenSearch back-end
 require_opensearch
 
+## Check for air gap deployment
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  source bin/airgap-include.sh
```
gsmith-sas, line 34: Why are we sourcing this file here; didn't we do that already in common.sh?

logging/bin/deploy_osd.sh
```diff
@@ -29,6 +29,22 @@ set -e
...
+
+  # Check for the image pull secret for the air gap environment
+  checkForAirgapSecretInNamespace "$AIRGAP_IMAGE_PULL_SECRET_NAME" "$LOG_NS"
+
+  # Check for the image pull secret for the air gap environment and replace placeholders
+  checkForAirgapSecretInNamespace "$AIRGAP_IMAGE_PULL_SECRET_NAME" "$LOG_NS"
```
gsmith-sas, line 40: Redundant call to checkForAirgapSecretInNamespace?

logging/bin/deploy_opensearch.sh
```diff
@@ -40,6 +40,20 @@ fi
 # Get/Set Helm Chart Version
 OPENSEARCH_HELM_CHART_VERSION=${OPENSEARCH_HELM_CHART_VERSION:-"2.11.0"}
 
+## Check for air gap deployment
+## Check for air gap deployment
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  source bin/airgap-include.sh
```
gsmith-sas, line 46: Why are we sourcing this file here; didn't we do that already in common.sh?

logging/bin/deploy_fluentbit_opensearch.sh
```diff
@@ -67,6 +67,18 @@ else
   log_debug "No existing release of the deprecated stable/fluent-bit Helm chart was found"
 fi
 
+## Check for air gap deployment
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  source bin/airgap-include.sh
```
gsmith-sas, line 72: Why are we sourcing this file here; didn't we do that already in common.sh?

logging/bin/deploy_fluentbit_azmonitor.sh
```diff
@@ -38,6 +38,19 @@ fi
 
 log_info "Deploying Fluent Bit (Azure Monitor)"
 
+## Check for air gap deployment
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  source bin/airgap-include.sh
```
gsmith-sas, line 43: Why are we sourcing this file here; didn't we do that already in common.sh?

gsmith-sas (changes requested):
Looking pretty good.  A few comments from original review unaddressed and few new things, but mostly minor-ish tweaks and clean-up.

logging/bin/deploy_eventrouter.sh
```diff
@@ -10,6 +10,21 @@ this_script=`basename "$0"`
...
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  
+  # Check for the image pull secret for the air gap environment
+  checkForAirgapSecretToNamespace "$AIRGAP_IMAGE_PULL_SECRET_NAME" "$LOG_NS"
+
+  eventrouterRegistry=$AIRGAP_REGISTRY
+  eventrouterPullSecret=$AIRGAP_IMAGE_PULL_SECRET_NAME
+fi
+
+echo "Event Router Registry: $eventrouterRegistry"
```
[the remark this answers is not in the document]
cumcke: Yep, that was debugging.  It'll be removed in the next push.

logging/bin/deploy_esexporter.sh
```diff
@@ -58,20 +58,48 @@ fi
...
 
+## Check for air gap deployment
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  
+  # Check for the image pull secret for the air gap environment
+  checkForAirgapSecretToNamespace "$AIRGAP_IMAGE_PULL_SECRET_NAME" "$LOG_NS"
+
+  # Copy template files to temp
+  airgapDir="$TMP_DIR/airgap"
+  mkdir -p $airgapDir
```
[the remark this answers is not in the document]
cumcke: I'll try moving it there again and making sure it works.  I was running into issues where, if it wasn't done at that time within the method, it kept referencing different directories and complaining about not being able to find the directory/file in question.
The change will be in the next push after I've tested it.

logging/bin/deploy_eventrouter.sh
```diff
@@ -10,6 +10,21 @@ this_script=`basename "$0"`
...
+eventrouterPullSecret="null"
+
+## Check for air gap deployment
+if [ "$AIRGAP_DEPLOYMENT" == "true" ]; then
+  
+  # Check for the image pull secret for the air gap environment
+  checkForAirgapSecretToNamespace "$AIRGAP_IMAGE_PULL_SECRET_NAME" "$LOG_NS"
+
+  eventrouterRegistry=$AIRGAP_REGISTRY
+  eventrouterPullSecret=$AIRGAP_IMAGE_PULL_SECRET_NAME
```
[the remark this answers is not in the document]
cumcke: Because the Event Router file is particular and doesn't have an additional airgap YAML file that we're adding onto it, it was the only way that I could think of to make sure that the registry and the pull secret were set to "gcr.io" and "null," respectively in the case where we were not doing an air gap deployment.

monitoring/airgap/cluster/airgap-values-prom-operator.yaml
```diff
@@ -0,0 +1,47 @@
...
+#  image:
+#    registry: __AIRGAP_REGISTRY__
+
+# =======================
+# Prometheus AlertManager
+# =======================
+# alertmanager:
+#  alertmanagerSpec:
+#    image:
+#      registry: __AIRGAP_REGISTRY__
```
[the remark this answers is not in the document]
cumcke: You're right.  This was more of a testing effort when I was running into an issue with the global setting.  Since I eventually got the global parameters working, we can delete the individual ones.  I'll fix that in my next push.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -47,14 +47,49 @@ if [ -z "$(kubectl get ns $MON_NS -o name 2>/dev/null)" ]; then
...
 
 set -e
 log_notice "Deploying monitoring to the [$MON_NS] namespace..."
 
 # Add the prometheus-community Helm repo
 helmRepoAdd prometheus-community https://prometheus-community.github.io/helm-charts
-log_debug "Updating Helm repositories..."
-helm repo update
+#log_debug "Updating Helm repositories..."
+#helm repo update
```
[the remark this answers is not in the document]
cumcke: Removed in the next push.

logging/bin/deploy_fluentbit_opensearch.sh
cumcke: Actually remembered to include the airgapValuesFile in this push.

bin/common.sh
cumcke: I have created a separate bin/airgap-include.sh file that is sourced in the deployment scripts.

logging/bin/deploy_opensearch.sh
gsmith-sas: Can ignore this comment.

logging/bin/deploy_fluentbit_opensearch.sh
gsmith-sas: Can ignore this comment.

logging/bin/deploy_fluentbit_opensearch.sh
gsmith-sas: wow!  I should have caught that...good catch!

logging/bin/deploy_eventrouter.sh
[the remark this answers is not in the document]
gsmith-sas: ok

PR #521, opened 2023-07-20
Addresses air gap related bugs and fixes

cumcke:


monitoring/airgap/airgap-grafana-values.yaml
```diff
@@ -0,0 +1,7 @@
+image:
+  repository: __AIRGAP_REGISTRY__/grafana/grafana
+  tag: "9.5.2"
```
gsmith-sas, line 3: We moved to 9.5.5 in #519.  So this will need to be updated here.

gsmith-sas (changes requested):
We upgraded Grafana due to a CVE (see #519); so you will need to update that one file.  Other than that, I think we're good.

gsmith-sas (approved):

PR #534, opened 2023-08-14
Fluentbit tracing

ceelias:


logging/fb/viya-tracing.conf
```diff
@@ -0,0 +1,12 @@
+# Version 1.9.9
+[INPUT]
+   name             opentelemetry
+   listen           0.0.0.0
+   port             4318
+   tag              viya_otel
+[OUTPUT]
+   Name opentelemetry
+   Host  v4m-tempo.monitoring
```
gsmith-sas, line 9: Hard-coded reference to "monitoring" namespace.  Can we just omit the namespace? Or, do we need to do some sort of substitution?

logging/fb/fluent-bit_config.configmap_opensearch.yaml
```diff
@@ -1038,7 +1038,12 @@ data:
        storage.sync normal
        storage.backlog.mem_limit 5M
        storage.metrics On
+
+    @INCLUDE viya-tracing.conf
```
gsmith-sas, line 1042: Fluent Bit doc says "_The @INCLUDE command only works at top-left level of the configuration line, it cannot be used inside sections._"  Based on that, I wonder if this would be better placed at the top rather than at the bottom.  Wouldn't impact functionality but would minimize chances of it getting "lost" after 1000 lines of log processing logic.  On the other hand, I suppose we plan on breaking this configuration into more manageable blocks/modules at some point; so maybe not worth changing for now.

logging/fb/viya-tracing.conf
```diff
@@ -0,0 +1,12 @@
+# Version 1.9.9
```
gsmith-sas, line 1: what does this version number represent?

gsmith-sas (changes requested):
The hard-coded reference to the 'monitoring' namespace is the only significant issue.  But take a look at the other comments as well.

logging/fb/fluent-bit_config.configmap_opensearch.yaml
ceelias: I have no problem moving this to the top but just to clarify, the "top" level the doc is referring is in reference to the indentation, meaning it can't be placed under `filter` or one of the other sections. But sure, def would make it more readable if it sits at the top

logging/fb/viya-tracing.conf
ceelias: I actually meant to ask you that because I copied this from viya-parsers.conf but I meant to delete it, which I will do. Thanks for catching it.

logging/fb/viya-tracing.conf
ceelias: I think I can remove, let me test this.

logging/fb/fluent-bit_config.configmap_opensearch.yaml
gsmith-sas: Understood.  My concern was mostly centered on making it obvious when we're including another file.  I also wonder about inclusion order and how duplicate/conflicting values are handled (i.e. are included files processed first and then the remaining content?  Or, are included files processed as they are encountered?).  We may want/need to dig into those messy details as we "modularize".

gsmith-sas (commented):
Let's disable the default value for TRACING_ENABLE.  Read/react to the other comments as appropriate.

gsmith-sas (approved):

PR #540, opened 2023-08-16
Agtar

gsmith-sas:


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # SAS Viya Monitoring for Kubernetes
 
+## Unreleased
+* **Overall**
+  * [FEATURE] Deploying into an air-gapped environment using local Helm archive (.tgz) files is now supported.  In this scenario,
+container images are still expected to be pulled from private container registry.  See
+[Configure SAS Viya Monitoring for Kubernetes for an Air-Gapped Environment](https://documentation.sas.com/?cdcId=obsrvcdc&cdcVersion=default&docsetId=obsrvdply&docsetTarget=n0grd8g2pkfglin12bzm3g1oik2p.htm).
+  * [CHANGE] The environment variables over-riding the default Helm chart versions to deploy MUST now be set in the user.env files
```
gsmith-sas, line 8: I have very mixed feelings about including this 2nd item in the CHANGELOG.  On the one hand, our move to the component_versions.env file does require/implement this change.  On the other hand, I don't believe we've documented this set of environment variables, so no one is likely using them.  Thoughts?

bin/helm-include.sh
```diff
@@ -95,8 +95,39 @@ function helmRepoAdd {
...
+     return 1
+  else
+     #all parms have values
+     :
+  fi
+
+  if [ "$AIRGAP_HELM_FORMAT"  == "tgz" ]; then
+      echo "${AIRGAP_HELM_REPO}${chart_name}-${chart_version}.tgz"
+  elif [ "$AIRGAP_HELM_FORMAT"  == "oci" ]; then
+      echo "oci://${AIRGAP_HELM_REPO}/${chart_repository}/${chart_name}"
```
gsmith-sas, line 122: I moved the prefixing with "oci://" to here and explicitly insert the "/" between the AIRGAP_HELM_REPO and the first part of the chart name.  While this can be considered a breaking change, I think the increased in usability outweigh the downsides, especially since the previous/existing functionality has only been available for a single release cycle.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -254,20 +258,24 @@ log_info "Deploying ServiceMonitors and Prometheus rules"
 log_verbose "Deploying cluster ServiceMonitors"
 
 if [ "$TRACING_ENABLE" == "true" ]; then
-  TEMPO_CHART_VERSION=${TEMPO_CHART_VERSION:-1.5.0}
   log_info "Tracing enabled..."
-  # # Add the grafana helm chart repo
+  # Add the grafana helm chart repo
```
gsmith-sas, line 262: Adding full air-gap support for Tempo is being handled in a separate work effort.

CHANGELOG.md
cumcke: We probably don't need to include the second bullet point - I don't think there's a good reason for users to change it and I wouldn't want someone to change it and cause their regular, online deployments to fail.

bin/helm-include.sh
cumcke: I agree.  I'm okay with this change, as well.  I don't think it will necessarily be breaking because the TGZ option would not have existed in the previous iteration and I would expect the deployment to work just fine.

bin/helm-include.sh
```diff
@@ -95,8 +95,39 @@ function helmRepoAdd {
...
+  elif [ -z "$chart_version" ]; then
+     echo "'ERROR: Helm chart version not specified'"
+     return 1
+  else
+     #all parms have values
+     :
+  fi
+
+  if [ "$AIRGAP_HELM_FORMAT"  == "tgz" ]; then
+      echo "${AIRGAP_HELM_REPO}/${chart_name}-${chart_version}.tgz"
```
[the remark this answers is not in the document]
cumcke: We should probably test to make sure that a file that matches the pattern exists in that spot before we get too far in the process.

cumcke (approved):

CHANGELOG.md
gsmith-sas: I have dropped the 2nd bullet point in my latest round of commits.

bin/helm-include.sh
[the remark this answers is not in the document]
gsmith-sas: I added a check in airgap-include.sh that confirms the AIRGAP_HELM_REPO is an existing directory when AIRGAP_HELM_FORMAT is "tgz".   That will trigger a failure very early in the deployment process if that environment variable isn't set properly.   I did NOT add a file exists check here because Helm actually already emits a good error message when that happens; adding our own would have been redundant.

PR #552, opened 2023-09-15
Getlogs 0.2.0

vishaanS:
COMPUTESVCS - 54012

- Changes:
- Added fields parameter to query generation, removing extraneous and unnecessary data.
- Refactored --out-file to make it much more efficient.
- Reading from environment variables is default method of connection.
- Replaced some created files with temporary ones to avoid clutter in working directory (Some files are user-created and will need to exist in CWD)
- Added protection against path traversal attacks in --out-file and --query-filename

### DIFF
--- a/logging/bin/getlogs.py
+++ b/logging/bin/getlogs.py
@@ -8,132 +8,66 @@ import argparse
 from opensearchpy import OpenSearch
 import os, sys
 import json, csv
+import tempfile
 
-## Cleanup temp file if not previously deleted, Create target directory
-DEFAULT_DIR = "GetLogsFiles"
-if not os.path.isdir(DEFAULT_DIR):
-    os.mkdir(DEFAULT_DIR)
-
-if os.path.isfile(DEFAULT_DIR + "/temp_SAS_OSQ.TXT"):
-    os.remove(DEFAULT_DIR + "/temp_SAS_OSQ.TXT")
+##v 0.2.0
 
 def validate_input(dict):
-    """Validate the arguments passed by the user to ensure script will function properly"""
-    
-    if(len(sys.argv) == 1): ##Check if any args have been provided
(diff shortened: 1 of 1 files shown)


logging/bin/getlogs.py
```diff
@@ -8,132 +8,66 @@
...
-            print(e)
-            exit()
-
-    if dict['format']:
-        if (dict['format'].lower() != 'csv') and (dict['format'].lower() != 'json'):
-            print("Error: Not a valid output type for --format. Supported filetypes are csv and json.")
-            exit()
-        else:
-            dict['format'] = dict['format'].lower()
+    ## Validate the arguments passed by the user to ensure script will function properly
```
ceelias, line 16: This should stay with 3 quotes as it is Python standard for doc string to explain purpose of a function, this is not a normal comment.

logging/bin/getlogs.py
```diff
@@ -153,52 +87,68 @@ def validate_input(dict):
...
+    if (dict['message']): ## Argument formatting for query builder
+        if(type(dict['message']) == list):
+            dict['message']= " ".join(dict['message'])
+        if (dict['message'].find("'") > -1): ##Check for invalid single quotes
+            print("Please remove single quotes ('') from search argument.")
+            exit()
+
 def build_query(dict): ##Generates Query using Opensearch DSL
-    """Takes query arguments from user and builds a query to pass to opensearch client"""
+    ##Takes arguments from user and builds a query to pass to opensearch client
```
ceelias, line 98: As explained above, this should stay with 3 quotes

ceelias (approved):
Looks good, just small comment about fixing docstrings comments for functions

ceelias (approved):
Looks good, just small comment about fixing docstrings comments for functions

ceelias (approved):

PR #557, opened 2023-09-25
update azure sample for tls by default

ceelias:


gsmith-sas (changes requested):
Looks good but I believe you are missing an annotation update

samples/azure-deployment/monitoring/user-values-prom-operator.yaml
gsmith-sas, line 1: I noticed the comment that starts on line 15 and ends on line 17 says that this sample is showing HTTP path-based ingress; which is no longer true.  Maybe reword the comment to something like: _"This example shows path-based ingress using HTTPS.  Refer to the [ingress samples](/samples/ingress) for other options."_  And keep the comment on line 18-19.

gsmith-sas (changes requested):
Looks good other than a comment in the user-values-prom-operator.yaml file that needs to be changed.

gsmith-sas (approved):
LGTM

Issue #549, opened 2023-09-14
sas-go-pods podmonitor incorrect


tynsh:
Hi,

we've had an issue with the sas-go-pods podmonitor. It tries to scrape all container ports exposed and not just the ones specified in the labels. Here is the relevant part from the prometheus targets page.

<html>
<body>

https://10.244.11.43:9999/internal/metrics | DOWN | container="sas-report-operations-core"endpoint="http"instance="10.244.11.43:9999"job="sas-report-operations"namespace="viya"node="aks-stateless-43847504-vmss000003"pod="sas-report-operations-7485dc55f6-fmmtb"sas_deployment="sas-viya"sas_service_base="golang"service="sas-report-operations" | 22.792s ago | 2.089ms | Get "https://10.244.11.43:9999/internal/metrics": http: server gave HTTP response to HTTPS client
-- | -- | -- | -- | -- | --
http://10.244.11.38:9114/internal/metrics | DOWN | container="sas-opendistro-exporter"endpoint="http"instance="10.244.11.38:9114"job="sas-opendistro-exporter"namespace="viya"node="aks-stateless-43847504-vmss000003"pod="sas-opendistro-exporter-5bcc7b79c5-nmp2f"sas_deployment="sas-viya"sas_service_base="golang"service="sas-opendistro-exporter" | 1.885s ago | 0.952ms | expected a valid start token, got "<" ("INVALID") while parsing: "<"
https://10.244.11.85:8080/internal/metrics | DOWN | container="sas-redis-operator"endpoint="http"instance="10.244.11.85:8080"job="sas-redis-operator"namespace="viya"node="aks-stateless-43847504-vmss000003"pod="sas-redis-operator-8487f55959-49zpr"sas_deployment="sas-viya"sas_service_base="golang"service="sas-redis-operator" | 23.228s ago | 1.359ms | Get "https://10.244.11.85:8080/internal/metrics": dial tcp 10.244.11.85:8080: connect: connection refused

</body>
</html>

We have added the following to the podMonitor spec:

```
    - action: replace
      regex: true;(.+)
      replacement: $1
      sourceLabels:
      - __meta_kubernetes_pod_annotationpresent_prometheus_io_port
      - __meta_kubernetes_pod_annotation_prometheus_io_port
      targetLabel: __port
    - action: replace
      regex: (.+);(.+)
      replacement: $1:$2
      sourceLabels:
      - __meta_kubernetes_pod_ip
      - __port
[...]

PR #559, opened 2023-09-28
[fix] update Go pod monitors to not scan unnecessary ports/pods

cumcke:


monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
```diff
@@ -22,11 +22,32 @@ spec:
     #   regex: golang
     #   action: keep
     - sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
-      regex: "sas-arke|sas-launcher"
+      regex: "sas-arke|sas-launcher|sas-redis-operator"
       action: drop
     - sourceLabels: [__meta_kubernetes_pod_annotation_sas_com_kustomize_base]
       regex: "golang"
       action: keep
+    - action: replace
```
cumcke, line 30: I don't know if it matters or not, but the order of the filter is "not like the others."
I'm not sure if this is a big deal or not, but I can go back and update them for consistency.

monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
```diff
@@ -22,11 +22,32 @@ spec:
     #   regex: golang
```
gsmith-sas, line 22: Take a look at the commented out lines on 19-23: would be good to determine whether most/all Go microservices are now using the correct value for that annotation...and then uncomment out the lines if appropriate.

monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
gsmith-sas: Yeah, I'm not crazy about having inconsistent ordering of things given how tough it often is to keep track of things in yaml files.  I think the issue is complicated  here by the fact that we were using sourceLabels as the "first" item in every set and these new relabelings include multiple sourceLabel values instead of just one.  I think tragetLabel might be a better "id"/first row since (I think) there is always only one target.  But I'm not sure if we need/want to reorder everything.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # SAS Viya Monitoring for Kubernetes
 
+## Unreleased
+
+* **Overall**
+
+* **Metrics**
+  * [FIX] Updated Go pod monitors to not scan unnecessary ports or pods.
```
gsmith-sas, line 8: Perhaps: [FIX] Updated Go podMonitors to eliminate scans of unnecessary ports or pods.

gsmith-sas (changes requested):
I think there some value in doing a bit more investigation on whether we can streamline this a bit and possible improve readability/usability.  See specific comments below.

monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
cumcke: Yeah, I just checked it out and that wasn't a change that I had made.  If it's already been implemented, I'll delete the comment.

monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
```diff
@@ -22,11 +22,32 @@ spec:
...
       action: drop
     - sourceLabels: [__meta_kubernetes_pod_annotation_sas_com_kustomize_base]
       regex: "golang"
       action: keep
+    - action: replace
+      regex: true;(.+)
+      replacement: $1
+      sourceLabels:
+      - __meta_kubernetes_pod_annotationpresent_prometheus_io_port
+      - __meta_kubernetes_pod_annotation_prometheus_io_port
```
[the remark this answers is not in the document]
cumcke: It turns out the annotationpresent one in this section is actually needed; if we don't have it, we get an error on 1-2 additional pods.  I'll reformat it to look like the rest of the file and update it in the next push.

monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
```diff
@@ -22,11 +22,32 @@ spec:
...
+      replacement: $1:$2
+      sourceLabels:
+      - __meta_kubernetes_pod_ip
+      - __port
+      targetLabel: __address__
+    - action: replace
+      regex: true;(.+)
+      replacement: $1
+      sourceLabels:
+      - __meta_kubernetes_pod_annotationpresent_prometheus_io_path
```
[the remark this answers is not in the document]
cumcke: I went ahead and removed this section because it does not affect the functionality of the scraping - I'm thinking it might've been a customization that they did for their own deployments.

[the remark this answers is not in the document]
gsmith-sas: I suspect this is related to the OpenSearch service running w/in SAS Viya.  The user is running an older version of SAS Viya.  In that version of SAS Viya, the OpenSearch instance w/in SAS Viya is monitored using an instance of the Elasticsearch Exporter...which makes metrics available on the endpoint /metrics rather than the /internal/metrics endpoint used by the SAS Golang-based pods.  

This block of code would tell the monitor to refer to the annotations to figure out the endpoint.  So, for the Elasticsearch Exporter pod, the annotations would tell it to use "/metrics" instead of "/internal/metrics".  Which is correct; and will likely eliminate this pod from showing up in Prometheus as "DOWN". So, problem solved!  EXCEPT the Elasticsearch Exporter isn't really one of the intended targets of this monitor (i.e. it is NOT an SAS-authored, Golang microservice).  Therefore,  I think we should handle this pod the same way we handle the other things that get "caught" by this monitor but we don't want to monitor; e.g. redis, sas-launcher, etc.  So, let's add "sas-opendistro-exporter" to that exclusion list.

monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
```diff
@@ -16,13 +16,18 @@ spec:
...
+      replacement: $1
+      targetLabel: __port
+    - sourceLabels: [__meta_kubernetes_pod_ip, __port]
+      action: replace
+      regex: (.+);(.+)
+      replacement: $1:$2
+      targetLabel: __address__
     - sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
-      regex: "sas-arke|sas-launcher"
+      regex: "sas-arke|sas-launcher|sas-redis-operator"
```
gsmith-sas, line 30: See my comment re: adding the sas-opendistro-exporter pod to the exclusion list.

gsmith-sas (changes requested):
See my comment re: adding the sas-opendistro-exporter pod to the exclusion list.

monitoring/monitors/viya/podMonitor-sas-go-pods.yaml
cumcke: Went ahead and added it to the exclusion list like we discussed.

gsmith-sas (approved):
Thanks for sticking with this and digging deeper into things.  LGTM.

gsmith-sas:
Resolved #549

PR #571, opened 2023-11-02
Fbevents prod

gsmith-sas:


logging/bin/deploy_fluentbit_k8sevents_opensearch.sh
```diff
@@ -0,0 +1,125 @@
...
+
+this_script=`basename "$0"`
+
+log_debug "Script [$this_script] has started [$(date)]"
+
+# Enable Fluent Bit?
+FLUENT_BIT_EVENTS_ENABLED=${FLUENT_BIT_EVENTS_ENABLED:-true}
+
+if [ "$FLUENT_BIT_EVENTS_ENABLED" != "true" ]; then
+   log_info "Environment variable [FLUENT_BIT_EVENTS_ENABLED] is not set to 'true'; existing WITHOUT deploying Fluent Bit deployment"
```
ceelias, line 18: `existing` ->  `exiting`

logging/bin/deploy_fluentbit_k8sevents_opensearch.sh
```diff
@@ -0,0 +1,125 @@
...
+source logging/bin/secrets-include.sh
+
+this_script=`basename "$0"`
+
+log_debug "Script [$this_script] has started [$(date)]"
+
+# Enable Fluent Bit?
+FLUENT_BIT_EVENTS_ENABLED=${FLUENT_BIT_EVENTS_ENABLED:-true}
+
+if [ "$FLUENT_BIT_EVENTS_ENABLED" != "true" ]; then
```
ceelias, line 17: I realize this is how we've been doing it but I feel as though this logic shouldn't live in this script. If I want to run this script I should be able to without having to also set the ENV variable. I think this logic should exist in any script that calls this one. 

I realize it defaults to true so I technically wouldn't have to but when I think about the reverse case, I wouldn't set this to false and then run this script. Maybe its not intended to be run as an individual script but I find it odd to have logic like this in the script itself. Not a big deal, may be personal preference, and may be missing some context but wanted to point it out.

ceelias (approved):
Looks solid, just one typo I found on log message so no big deal

logging/bin/deploy_fluentbit_k8sevents_opensearch.sh
gsmith-sas: Good catch...Corrected.

PR #573, opened 2023-11-10
Getlogs Port Forwarding Functionality 

vishaanS:


logging/bin/getlogs.py
```diff
@@ -95,6 +103,21 @@ def validate_input(dict):
...
+def open_port(dict):
+    """Binds the v4m-search service on port 9200 to a locally available port"""
+    #Get open port
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.bind(("",0))
+    s.listen(1)
+    port = s.getsockname()[1]
+    s.close()
+
+    cmd = (["kubectl", "-n", "logging", "port-forward", "svc/v4m-search", str(port) + ':9200', '&'])
```
gsmith-sas, line 115: This contains a  hard-coded reference to the "logging" namespace..  While that is the default name for the namespace in which the Viya 4 Monitoring log monitoring components are deployed, users can use something else.  Therefore, this script should find the namespace in which the v4m-search service is deployed and use it here.  Sorry, I suspect that detail was probably left out of the requirements.

In addition, what happens if this command fails?  Do we need logic detecting the failure?  Or, will it fail in a way that makes it clear to the user what happened?  I suspect it doesn't.

logging/bin/getlogs.py
```diff
@@ -197,6 +220,7 @@ def get_arguments():
     
     parser.add_argument('-i', '--index', required=False, dest="index", metavar="INDEX", default="viya_logs-*") ## help = "\nDetermine which index to perform the search in. Default: viya-logs-*\n\n
     ##Connection settings
+    parser.add_argument('-pf','--port-forward', required=False, dest="portforward", action="store_true", help = "\n If this option is provided, getlogs will use the value in your KUBECONFIG (case-sensitive) environment variable to port-forward and connect to the open-search API. This skips ESHOST and ESPORT, but ESUSER and ESPASSWD are stil required to authenticate and connect to the database. \n\n")
```
gsmith-sas, line 223: "OpenSearch" not "open-search"

logging/bin/getlogs.py
```diff
@@ -95,6 +103,21 @@ def validate_input(dict):
             print("Please remove single quotes ('') from search argument.")
             sys.exit()
 
+def open_port(dict):
+    """Binds the v4m-search service on port 9200 to a locally available port"""
+    #Get open port
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.bind(("",0))
+    s.listen(1)
+    port = s.getsockname()[1]
```
gsmith-sas, line 112: Does this reserve/lock the port preventing anything else from using it?  If not, I suspect there is a very small chance that the port identified here is taken by some other process before the kubectl command executes.  Probably not worth worrying about but a potential issue.

The kubectl port-forward command will grab a random port if none is provided but then you have to figure out which port was used...that may be difficult/messy.

gsmith-sas (changes requested):
See comments for details but this code includes a hard-coded reference to the namespace and lacks error checking.  Both seem "significant".

logging/bin/getlogs.py
ceelias: The `bind` command above will bind the socket to any open port which should take the risk you mention out of question. Once the socket binds it should own the port

logging/bin/getlogs.py
```diff
@@ -95,6 +103,21 @@ def validate_input(dict):
             print("Please remove single quotes ('') from search argument.")
             sys.exit()
 
+def open_port(dict):
```
ceelias, line 106: I know some of the has already done but I want to point out that I'm not a fan of passing around dictionaries to functions like this. I think arguments should be passed explicitly. You could define this function with the arguments such as:
`def open_port(host, port)`
and then pass the dictionary, which still isn't great but like this:

`open_port(**dict)`
That should resolve the dictionary to the arguments of the function.

The reason I prefer explicit arguments is that it makes it much easier to maintain and readable.

logging/bin/getlogs.py
```diff
@@ -206,6 +230,8 @@ def get_arguments():
 
 args = get_arguments() ##Creates "args" dictionary that contains all user submitted options. Print "args" to debug values. Note that the 'dest' value for each argument in argparser object is its key.
 validate_input(args)
+if args['portforward']:
+    open_port(args)
```
ceelias, line 234: Should we just pass `args['portforward']`? Similar to my comment above, it can get dangerous just passing dictionaries around. The minimal amount of arguments necessary should be passed.

PR #579, opened 2023-11-16
Disable init-chown-data initContainer for Grafana

cumcke:


cumcke:
> All of this looks fine. Just in case someone needs to revert to the original behavior, I think we should add the initChownData stanza to the user-values-prom-operator.yaml files in monitoring and samples/generic-base/monitoring as well. That Helm chart's user yaml file is several thousand lines long and this particular value is actually part of the Grafana sub-chart: I think making a user who wants to revert to the original behavior dig through all of that would be user-hostile. Leaving a few "breadcrumbs" in those files to help them find their way would be nicer.

I went ahead and added the parameter to the user-values-prom-operator.yaml file and the generic base sample.

monitoring/user-values-prom-operator.yaml
```diff
@@ -64,6 +64,8 @@
 #   enabled: true
 #   adminPassword: yourSecretPasswordHere
 #   "grafana.ini":
+#     initChownData:
```
gsmith-sas, line 67: Is initChownData part of the "grafana.ini" file?  The indentation here makes that look like the case.  However, if I look at the Grafana Helm chart, it looks like the initChownData stanza is a top-level item (i.e. w/no indentation)...which would mean it should be indented one-level below the grafana: tag here in this chart.

monitoring/values-prom-operator.yaml
```diff
@@ -196,6 +196,9 @@ grafana:
         format: json
     dashboards:
       default_home_dashboard_path: /tmp/dashboards/viya-welcome-dashboard.json
+    initChownData:
```
gsmith-sas, line 199: Hmmm, still concerned that the indentation might not be exactly right.  In this file,  it appears that the **intiChownData:** line lines up with the **analytics:**, **log:,** **"log.console":** and **dashboards:** lines...which all seem to be indented two spaces deeper than the **image:** and **"grafana.ini":** lines.  I suspect those should all line up together.  As they are now, it makes them all appear to be "children" under the **"grafana.ini":** line.  Note that they lineup with the **image.tag** line.

samples/generic-base/monitoring/user-values-prom-operator.yaml
```diff
@@ -64,6 +64,8 @@
 #   enabled: true
 #   adminPassword: yourSecretPasswordHere
 #   "grafana.ini":
+#      initChownData:
+#       enabled: false
```
gsmith-sas, line 68: See comments above.

gsmith-sas (commented):
@cumcke   Please double-check the indentation levels within these files; as noted in my comments, somethings don't look exactly right.

monitoring/user-values-prom-operator.yaml
cumcke: I moved the parameter and it is working as expected now.

samples/generic-base/monitoring/user-values-prom-operator.yaml
```diff
@@ -52,20 +52,20 @@
...
 # alertmanager:
-#   enabled:true
+#   enabled: true
 #   alertmanagerSpec:
 #     externalUrl: http://host.mycluster.example.com:31091
-#     storage:
-#       volumeClaimTemplate:
-#         spec:
-#           storageClassName: myCustomStorageClass
+#   storage:
```
gsmith-sas, line 58: I think storage is a "child" of alertmanagerSpec and NOT of alertmanager.  That is how we are handling it in monitoring/values-prom-operator.yaml  (see [line 138 in that file](https://github.com/sassoftware/viya4-monitoring-kubernetes/blob/7fe23a8157f666c1c482dd2d13e88794745c58d6/monitoring/values-prom-operator.yaml#L138)).    And/or see [line 724 in values.yaml file in the Helm chart](https://github.com/prometheus-community/helm-charts/blob/5b2ac2b7c404bc14fe1bb42a5f96a4a45d7333f7/charts/kube-prometheus-stack/values.yaml#L724)

In addition, samples/generic-base/monitoring/user-values-prom-operator.yaml should generally match the contents of monitoring/values-prom-operator.yaml.  So, any change in one of those files should generally also be needed in the other one.  I notice you did not make this change in both files.

gsmith-sas (changes requested):
I think the initChown changes are fine now...but it looks like you may have introduced a problem related to the alertmanagerSpec.storage stanza...see notes below.

gsmith-sas (approved):
LGTM

PR #589, opened 2023-12-09
[feature] Added SAS Arke dashboard to Grafana

cumcke:


monitoring/dashboards/viya/sas-arke-dashboard.json
```diff
@@ -0,0 +1,503 @@
...
+    "fiscalYearStartMonth": 0,
+    "graphTooltip": 0,
+    "id": 58,
+    "links": [],
+    "liveNow": false,
+    "panels": [
+      {
+        "datasource": {
+          "type": "prometheus",
+          "uid": "PBFA97CFB590B2093"
```
gsmith-sas, line 29: Don't we have to change these to some generic datasource reference rather than this specific one?  a quick look at a few of our existing dashboards seems to show references like `"datasource": "Prometheus"`. Without making this change, I believe you will get error messages like: "
```
Templating Failed to upgrade legacy queries 
Datasource PBFA97CFB590B2093 was not found
```

monitoring/dashboards/viya/sas-arke-dashboard.json
```diff
@@ -0,0 +1,503 @@
...
+    "tags": [
+      "sas-viya"
+    ],
+    "templating": {
+      "list": [
+        {
+          "current": {
+            "selected": false,
+            "text": "d69195",
+            "value": "d69195"
```
gsmith-sas, line 468: This is a references to the namespace for the SAS Viya deployment in on on my test clusters; pretty sure it needs to be made more "generic" as well.  Ideally it would default to the namespace of a SAS Viya deployment that exists in the cluster or, if that's not easily done,  perhaps have no default value.

gsmith-sas (changes requested):
I think you need to make the datasource values w/in the JSON to make them "generic".  See notes below.

monitoring/dashboards/viya/sas-arke-dashboard.json
cumcke: I went ahead and removed the whole "current" parameter so the namespace reference would be good.  It has been addressed in this latest push.

monitoring/dashboards/viya/sas-arke-dashboard.json
cumcke: Yep, you are right.  I think this was a byproduct of how the dashboard was originally saved.
I have updated those references to be generic **$datasource** like our other dashboards.
It has been updated in the latest push.

gsmith-sas (approved):

cumcke (commented):
Looks good.  I know some of the existing references to the data source use "$datasource," but I think "Prometheus" should be fine.

PR #590, opened 2023-12-11
Tempo node graph

ceelias:


monitoring/values-prom-operator.yaml
```diff
@@ -237,3 +237,11 @@ grafana:
       - kubernetes.io/pvc-protection
     # subPath: ""
     # existingClaim:
+
+tempo: 
+  metricsGenerator:
+    enabled: true
```
gsmith-sas, line 243: What happens if Tempo is not enabled: will this setting cause any problems?

monitoring/values-prom-operator.yaml
```diff
@@ -237,3 +237,11 @@ grafana:
       - kubernetes.io/pvc-protection
     # subPath: ""
     # existingClaim:
+
+tempo:
```
gsmith-sas, line 241: I don't see a tempo: stanza available in the [kube-prometheus-stack values.yaml](https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml) file nor in the [grafana values.yaml](https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml) file.  Should this be in the values-tempo.yaml file?

gsmith-sas (commented):
Please see my question about where the updated tempo: stanza goes...I think it may be in the wrong file.

monitoring/values-tempo.yaml
```diff
@@ -13,8 +13,12 @@ tempo:
   reportingEnabled: false
   metricsGenerator:
     # -- If true, enables Tempo's metrics generator (https://grafana.com/docs/tempo/next/metrics-generator/)
-    enabled: false
-    remoteWriteUrl: "http://prometheus.monitoring:9090/api/v1/write"
+    enabled: true
+    remoteWriteUrl: "http://v4m-prometheus.monitoring:9090/api/v1/write"
```
gsmith-sas, line 17: Hmmm, this contains a hard-coded namespace reference.  Is there any way to _easily_ make this "generic" and avoid that?  If not, let's leave it as-is for now but plan on fixing it in an upcoming sprint.  If someone bumps into it before then we can walk them through providing the correct value in their USER_DIR/monitoring/user-values-tempo.yaml file.

gsmith-sas (approved):
Looks good.

PR #592, opened 2023-12-12
adding github action to check shell code

ceelias:
Look at #593 to see what the result of this action looks like.


gsmith-sas (approved):
LGTM

PR #598, opened 2023-12-19
Production changes

vishaanS:
Final changes for getlogs - no functionality changes just code style updates


ceelias (approved):

gsmith-sas (changes requested):
I found one typo and a number of other things that need to be tweaked in the usage output.  I was able to add a comment for most of them in-line but for some I could not.  

Here is information about those:
- In the usage information, please remove the line starting with "Additional arguments..." since that is not the complete list of fields.
- It appears that the ID field is always included in the output.  Therefore, it should be included in the list of default fields AND we should probably mention that the field is always included (as the last column).

gsmith-sas (approved):
LGTM

PR #644, opened 2024-05-14
Fix to support removal of --short option from kubectl version

greg-wootton:
kubectl version --short is no longer supported at 1.28. Switching to using --output=json and parsing this with tr and sed.

### DIFF
--- a/bin/kube-include.sh
+++ b/bin/kube-include.sh
@@ -11,8 +11,8 @@ if [ ! "$(which kubectl)" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


gsmith-sas:
@greg-wootton Thank you for this PR.  Unfortunately, the project team is a little underhanded at the moment due to illness and other unavoidable absences.  However, we hope to operational again by next week.  At that point, we will review this PR and sort out a plan for incorporating it into project.  We'll keep you updated.

Regards,
Greg

gsmith-sas:
We decided to handle this issue with slightly different syntax (pushed as #652).  Abandoning this PR.

PR #652, opened 2024-06-12
[CHANGE] Eliminated use of  --short option from kubectl version commands

gsmith-sas:


gsmith-sas:
This PR supersedes/ closes #644 .

PR #684, opened 2024-10-07
[SECURITY] Disable KSM Metrics related to K8s Secret resources

gsmith-sas:


gsmith-sas:
The v4m-kube-state-metrics ClusterRole was granted 'list' and 'watch' permission for Kubernetes Secrets resources.  Having access to secrets enables an account to potentially escalate privileges. By retrieving a secret, an attacker would be able to access the Kubernetes API Server in the context of that account.  We considered whether granting these permissions were strictly necessary and weighed the impact of removing them.  The project does not currently make use of the metrics about Kubernetes Secret resources (i.e. kube_secret_info, kube_secret_type, kube_secret_labels and kube_secret_created).  Therefore, following the principal of least privilege, we decided that there was no compelling reason to continue to grant these permission.

Users who need/want to re-enable these metrics after considering the security implications can do so by making the changes to the user-values-prom-operator.yaml file in their $USER_DIR/monitoring sub-directory prior to (re)deploying the metric monitoring components.

PR #712, opened 2025-01-27
Changed delimiter for sed w/in v4m_replace function

gsmith-sas:


gsmith-sas:
Not sure if we need/want to list this in the changelog or not.  I was concerned the comment might raise more questions.  Let me know what you think.

bin/common.sh
```diff
@@ -279,9 +279,9 @@ function parseFullImage {
 function v4m_replace {
 
     if echo "$OSTYPE" | grep 'darwin' > /dev/null 2>&1; then
-      sed -i '' s/"$1"/"$2"/g  "$3"
+      sed -i '' "s;$1;$2;g"  "$3"
```
ceelias, line 282: I’m a little worried these delimiters won’t work properly on OSX, sed is always iffy between the different OSs. Whats an easy way for me to verify this works on my machkne?

ceelias:
> Not sure if we need/want to list this in the changelog or not. I was concerned the comment might raise more questions. Let me know what you think.

I think it makes sense to include in the changelong still

bin/common.sh
ephesused: `sed` has long-running support for other delimiters, but an easy way to confirm on your machine is to check the man page. On my mac:

````
     [2addr]s/regular expression/replacement/flags
             Substitute the replacement string for the first instance of the regular expression in the pattern
             space.  Any character other than backslash or newline can be used instead of a slash to delimit
             the RE and the replacement.  Within the RE and the replacement, the RE delimiter itself can be
             used as a literal character if it is preceded by a backslash.
````

ceelias: reading man pages makes me sad 😓 , i prefer to test it  😆  I appreciate the input!!

I tested and can doubly confirm things work

ceelias (approved):

PR #740, opened 2025-04-14
Shellcheck fluentbit scripts

ceelias:


logging/bin/deploy_fluentbit_azmonitor.sh
```diff
@@ -1,46 +1,55 @@
-#! /bin/bash
+#!/bin/bash
 
 # Copyright © 2020, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-cd "$(dirname $BASH_SOURCE)/../.."
+# shellcheck source=/dev/null
+cd "$(dirname "$BASH_SOURCE")/../.." || exit 1
```
ceelias, line 7: I looked into this quoting. Apparently bash knows the resolve the variable resolution and the quoting inside the quotes dont break anything.

logging/bin/deploy_fluentbit_azmonitor.sh
```diff
@@ -1,46 +1,55 @@
-#! /bin/bash
+#!/bin/bash
 
 # Copyright © 2020, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-cd "$(dirname $BASH_SOURCE)/../.."
+# shellcheck source=/dev/null
```
gsmith-sas, line 6: What's this line do?

logging/bin/deploy_fluentbit_k8sevents_opensearch.sh
```diff
@@ -75,43 +79,41 @@ if [ ! -f "$FB_EVENTS_USER_YAML" ]; then
   FB_EVENTS_USER_YAML=$TMP_DIR/empty.yaml
 fi
 
-
 # Point to OpenShift response file or dummy as appropriate
+openshiftValuesFile="$TMP_DIR/empty.yaml"
```
gsmith-sas, line 83: Human-made change or shellcheck?

logging/bin/deploy_fluentbit_opensearch.sh
```diff
@@ -110,104 +113,103 @@ else
...
-    containerd|cri-o)
-      KUBERNETES_RUNTIME_LOGFMT="criwithlog"
-      ;;
-    *)
-      log_warn "Unrecognized Kubernetes container runtime [$runtime]; using default parser"
-      KUBERNETES_RUNTIME_LOGFMT="docker"
-      ;;
-   esac
+  somenode=$(kubectl get nodes | awk 'NR==2 { print $1 }')
+  # Quote the jsonpath expression to avoid SC1083
```
gsmith-sas, line 141: You comment or shellcheck's?

logging/bin/deploy_fluentbit_azmonitor.sh
```diff
@@ -106,94 +122,96 @@ else
...
-  kubectl -n "$LOG_NS" create configmap fbaz-viya-tracing  --from-file="$TMP_DIR"/empty.yaml
+  kubectl -n "$LOG_NS" create configmap fbaz-viya-tracing --from-file="$TMP_DIR"/empty.yaml
 
-  tracingValuesFile=$TMP_DIR/empty.yaml
+  tracingValuesFile="$TMP_DIR/empty.yaml"
 fi
 
 # Check for Kubernetes container runtime log format info
-KUBERNETES_RUNTIME_LOGFMT="${KUBERNETES_RUNTIME_LOGFMT}"
+KUBERNETES_RUNTIME_LOGFMT="${KUBERNETES_RUNTIME_LOGFMT:-}"
```
gsmith-sas, line 148: Sets default value to null?

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SHELLCHECK_OPTS: -e SC1004 # exclude some shellcheck warnings.
-          SHFMT_OPTS: -s # arguments to shfmt.
+          SHFMT_OPTS: -s -i 2 -bn -sr -ln bash # shfmt: simplify (s), 2-space indentation (-i 2), braces on same line (-bn), space after redirects (-sr), bash dialect (-ln bash)
```
gsmith-sas, line 14: As I recall, you're preference was indenting 4 spaces, mine was 3 spaces and shellcheck's default was 12.  So, how did we end up with 2? ;-)  I think 2 is too little and 12 is way too much.

gsmith-sas (commented):
I don't see any problems but did have some questions...will approve after discussing.

logging/bin/remove_fluentbit_k8sevents_opensearch.sh
```diff
@@ -3,15 +3,18 @@
 # Copyright © 2023, SAS Institute Inc., Cary, NC, USA.  All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-cd "$(dirname $BASH_SOURCE)/../.."
+cd "$(dirname "$BASH_SOURCE")/../.." || {
+  echo "Failed to change directory"
+  exit 1
+}
```
gsmith-sas, line 9: Why is this being handled differently in different scripts?  Here is echos a message when failing to change directory, in other places it doesn't.  Here it keeps the $BASH_SOURCE reference but in some places (e.g. logging/bin/remove_fluentbit_azmonitor.sh) it changes the reference to "$0"

logging/bin/deploy_fluentbit_azmonitor.sh
ceelias: it was an original ref to make spellcheck ignore the source but I ignored it globally. I removed comment.

logging/bin/deploy_fluentbit_azmonitor.sh
ceelias: This was recommended by shellcheck, but I added it. It basically will exit the script if the "cd" command fails to execute.

bin/helm-include.sh
```diff
@@ -128,7 +128,8 @@ function get_helm_versionstring {
     if [ "$V4M_HELM_USE_LATEST"  == "true" ]; then
        :  # return null string
     else
-       echo "--version $1"
+       # Explicitly use printf instead of echo to avoid quote issues
+       printf -- "--version %s" "$1"
```
ceelias, line 132: basically the issue we were seeing was that "echo" preserves the quotes when this command is called. so that's why were were getting double quotes. with printf, we should not have this issue.

logging/bin/deploy_fluentbit_opensearch.sh
ceelias: mine. I removed it now

.github/workflows/ci.yml
ceelias: changed to 4

logging/bin/deploy_fluentbit_azmonitor.sh
ceelias: it actually sets it to empty string, which I made sure doesn't break the if block on the next line. This is recommended by shellcheck as its considered more robust

gsmith-sas (approved):
looks good.

PR #745, opened 2025-04-22
Shellcheck mon scripts

ceelias:


monitoring/bin/change_grafana_admin_password.sh
```diff
@@ -3,140 +3,145 @@
...
 
 # Changes the admin password using the Grafana CLI
 log_info "Updating the admin password using Grafana CLI"
-kubectl exec -n $namespace $grafanaPod -c grafana -- bin/grafana-cli admin reset-admin-password $password
-
-# Exit out of the script if the Grafana CLI call fails
-if (( $? != 0 )); then
-  log_error "An error occurred when updating the password"
-  exit 1
+if ! kubectl exec -n "$namespace" "$grafanaPod" -c grafana -- bin/grafana-cli admin reset-admin-password "$password"; then
```
gsmith-sas, line 127: Interesting that this is preferred...definitely looks a bit odd.

monitoring/bin/common.sh
```diff
@@ -1,42 +1,44 @@
...
+    export MON_NS="${MON_NS:-monitoring}"
+    export TLS_ENABLE="${MON_TLS_ENABLE:-${TLS_ENABLE:-true}}"
 
-     #Special processing to handle Viya-level deployment script
-     if [ `basename "$0"` == "deploy_monitoring_viya.sh" ]; then
-        V4M_NS="$VIYA_NS"
-     else
-        export V4M_NS=$MON_NS
-     fi
+    if [ "$AIRGAP_DEPLOYMENT" = "true" ]; then
```
gsmith-sas, line 24: Why the change from "==" to "=" here but not elsewhere?

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -96,400 +95,398 @@ helmRepoAdd prometheus-community https://prometheus-community.github.io/helm-cha
...
+    # shellcheck disable=SC2016
+    yq -i eval-all '. as $item ireduce ({}; . * $item )' "$autogeneratedYAMLFile" "$ingressSampleFile"
+
+    ALERTMANAGER_INGRESS_ENABLE="${ALERTMANAGER_INGRESS_ENABLE:-false}"
+
+    amIngressCert="${ALERTMANAGER_INGRESS_CERT}"
+    amIngressKey="${ALERTMANAGER_INGRESS_KEY}"
+    create_ingress_certs "$MON_NS" "alertmanager-ingress-tls-secret" "$amIngressCert" "$amIngressKey"
+
+    # Don't assign the variable to itself
```
gsmith-sas, line 209: Comment?  Is this a reminder to make a change?
We used this approach to ensure the variable is exists and is set, i.e. if set, use the set value;  if not set, create it and use the default value. 
What's preferred alternative?

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -96,400 +95,398 @@ helmRepoAdd prometheus-community https://prometheus-community.github.io/helm-cha
...
+        fi
+    fi
+
+    GRAFANA_INGRESS_ENABLE="${GRAFANA_INGRESS_ENABLE:-true}"
+
+    grIngressCert="${GRAFANA_INGRESS_CERT}"
+    grIngressKey="${GRAFANA_INGRESS_KEY}"
+    create_ingress_certs "$MON_NS" "grafana-ingress-tls-secret" "$grIngressCert" "$grIngressKey"
+
+    # Don't assign the variable to itself
```
gsmith-sas, line 225: Comment?  See above.

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -96,400 +95,398 @@ helmRepoAdd prometheus-community https://prometheus-community.github.io/helm-cha
...
+            GRAFANA_FQDN="$BASE_DOMAIN"
+        fi
+    fi
+
+    PROMETHEUS_INGRESS_ENABLE="${PROMETHEUS_INGRESS_ENABLE:-false}"
+    prIngressCert="${PROMETHEUS_INGRESS_CERT}"
+    prIngressKey="${PROMETHEUS_INGRESS_KEY}"
+    create_ingress_certs "$MON_NS" "prometheus-ingress-tls-secret" "$prIngressCert" "$prIngressKey"
+
+    # Don't assign the variable to itself
```
gsmith-sas, line 240: Comment? See above.

monitoring/bin/deploy_dashboards.sh
```diff
@@ -25,108 +25,108 @@ TEST_DASH="${TEST_DASH:-false}"
...
-       log_debug "Deploying dashboard from file [$f]"
-       name=$(basename $f .json)
-       
-       kubectl create cm -n $DASH_NS $name --dry-run=client --from-file $f -o yaml | kubectl apply --server-side -f -
-       kubectl label cm -n $DASH_NS $name --overwrite grafana_dashboard=1 sas.com/monitoring-base=kube-viya-monitoring sas.com/dashboardType=$type
-     fi
-   done
+    type=$1
+    if [ -z "$2" ]; then
+        dir="$DASH_BASE/$type"
```
gsmith-sas, line 30: Interesting.  It appears that calling this function with a single argument (which corresponds to a sub-directory under $DASH_BASE _only_ works from _within_ this script.  Calling the script (rather than the function) with a single argument will fail if the argument does not refer to a full directory (due to the validation around line 67).  

I'm not sure that's optimal but I guess this is how things have worked for a very long time, so we leave it as-is.

monitoring/bin/deploy_monitoring_openshift.sh
```diff
@@ -40,325 +40,324 @@ helmRepoAdd grafana https://grafana.github.io/helm-charts
...
+    log_debug "Annotating grafana-serviceaccount to auto-generate TLS certs..."
+    kubectl annotate serviceaccount -n "$MON_NS" --overwrite grafana-serviceaccount 'serviceaccounts.openshift.io/oauth-redirectreference.primary={"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"v4m-grafana"}}'
+
+    log_debug "Adding ClusterRoleBinding for grafana-serviceaccount..."
+    crbYAML=$TMP_DIR/grafana-serviceaccount-binding.yaml
+    cp monitoring/openshift/grafana-serviceaccount-binding.yaml "$crbYAML"
+    if echo "$OSTYPE" | grep 'darwin' > /dev/null 2>&1; then
+        sed -i '' "s/__MON_NS__/$MON_NS/g" "$crbYAML"
+    else
+        sed -i "s/__MON_NS__/$MON_NS/g" "$crbYAML"
```
gsmith-sas, line 233: Replace lines 230-234 with call to the v4m_replace function: i.e. `v4m_replace "__MON_NS__" "$MON_NS" "$crbYAML"`?

gsmith-sas (changes requested):
Mostly questions/clarifications...but a couple of changes too.

monitoring/bin/common.sh
ceelias: I actually decided to go back to "==". I think since we are using bash specifically, its ok for us to use it.

monitoring/bin/change_grafana_admin_password.sh
ceelias: I actually prefer the way it looks but if you don't we can change it back, I'm not that stuck on it

monitoring/bin/deploy_monitoring_cluster.sh
ceelias: Yea it was just a comment to remove that because it was a useless line. I'll remove the comments

monitoring/bin/deploy_dashboards.sh
ceelias: Yea, that's a good point. Think it's worth making a ticket to address this later? Or worth me looking for a solution now?

monitoring/bin/change_grafana_admin_password.sh
gsmith-sas: No, I don't think there's a need to revert this, I can get used to it.  I was going to ask about whether we need/want to redirect stderr/stdout but I guess any messages generated by the command could be useful...and, as I recall, since it is part of an IF statement, the exit-on-error (set -e) processing won't kick in.

monitoring/bin/deploy_dashboards.sh
gsmith-sas: My initial reaction was 'no, I think we just accept it as a "quirk"'.  But then I checked the documentation and we do say it can be used to deploy a single dashboard.  I don't think this is critical, so no need to address it now; just add a story to the backlog.

.shellcheckrc
```diff
@@ -2,5 +2,5 @@
 disable=SC1091
 # Disables the check for expanding an array, as it predictably gives first element
 disable=SC2128
-# Disbale check for var not set in same file
+# Disabales check for var not set in same file
```
gsmith-sas, line 5: Almost correct ;-) ."Disbale" => "Disabales"...should be "Disables".

gsmith-sas (changes requested):
One small typo, but other looks good.  Fix the typo and I think this is good to go.

.shellcheckrc
ceelias: 😆 my brain clearly wasn't working

gsmith-sas (approved):
Looks good to me.

PR #746, opened 2025-04-22
Remove configuration-snippet annotation from Ingress definitions

gsmith-sas:


gsmith-sas:
Unfortunately, these changes will not work with ingress-nginx 1.12+ unless/until the OpenSearch Helm charts support setting the **pathType** to `ImplementationSpecific`.  There is a OpenSearch Pull Request ([PR 471](https://github.com/opensearch-project/helm-charts/pull/471)) to support this enhancement but it appears stagnant at this time.  Abandoning this PR (for now) and will try again if/when the OpenSearch Helm charts are updated.

PR #755, opened 2025-05-22
Removemigrateshellck

gsmith-sas:


logging/bin/secrets-include.sh
```diff
@@ -8,101 +8,105 @@
...
+            return 9
+        fi
+
+        if [ "$(kubectl -n "$LOG_NS" create secret generic "$secret_name" --from-file="$filepath"/"$file")" == "secret/$secret_name created" ]; then
+            log_verbose "Created secret for OpenSearch config file [$file]"
+
+            if [ -n "$label" ]; then
+                log_debug "Applying label [$label] to newly created secret [$secret_name]"
+                # shellcheck disable=SC2086
+                # quoting $label results in invalid kubectl syntax
```
gsmith-sas, line 34: I wonder if we should have _require_ an explanation when disabling any shellcheck directive in the code?  Adding an explanation after the disable directive (e.g. `#shellcheck disable=SC2086 quoting $label results in invalid kubectl syntax`) is NOT valid, so I had to add a 2nd comment.   The Shellcheck doc shows including both comments on the same line as acceptable alternative, i.e. `#shellcheck disable=SC2086    #quoting $label results in invalid kubectl syntax`.  That might be cleaner/more readable.  Thoughts?

PR #759, opened 2025-06-16
Grafana alerts

ceelias:


monitoring/values-prom-operator.yaml
```diff
@@ -285,4 +279,10 @@ grafana:
       - kubernetes.io/pvc-protection
     # subPath: ""
     # existingClaim:
-
+  extraVolumes:
```
gsmith-sas, line 282: Isn't the creation of the ConfigMap optional?  If $INCLUDED_ALERTS_DIR is not set, the configMap is not created.  If the configMap is not created, won't specifying it as an extra Volume cause the pod to fail to start?  Easiest fix might be to always create the configMap even if it is an "empty" file.  Or, only add the extra volume when necessary.

monitoring/alerting/rules/sas-job-launcher-rules.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: 1
```
gsmith-sas, line 1: "Provisioned" resources cannot be modified by users via Grafana.  I think that's ok for dashboards since users can clone an existing dashboard, tweak it and save it with a new name.  But for alerts, this might not be optimal since the original alert will still be there and potentially trigger at the wrong time.  But we need/want to provide some OOTB alerts, so it's a tough call.

monitoring/alerting/rules/viya-alert-rules.yaml
```diff
@@ -0,0 +1,1236 @@
...
+            refId: C
+            relativeTimeRange:
+              from: 600
+              to: 0
+        execErrState: Error
+        for: 5m
+        isPaused: false
+        labels: {}
+        noDataState: NoData
+        title: cas-restart
```
gsmith-sas, line 97: How does the "cas-restart" Alert differ from the "CAS Restart Alert" alert?  The names seem very similar and the Summary information are identical.  Are they duplicates?

Same question for the "cas-memory" vs. "CAS Memory Usage" alerts?    See similar things for some of the Crunchy-related alerts as well.

gsmith-sas (commented):
I think we need to talk through some of these things.

monitoring/values-prom-operator.yaml
ceelias: It is not optional, it will be created based on the provided ones we have. If they add more, those will just be added to the existing configmap but the configmap will always exist.

monitoring/alerting/rules/database_alerts.yaml
```diff
@@ -0,0 +1,212 @@
...
+    orgId: 1
+    rules:
+      - title: Catalog DB Connections High
+        annotations:
+          description:
+            "Checks the in-use catalog database connections > 21. The default
+            db connection pool is 22.  If it reaches the limit, the rabbitmq queues starts
+            to fill up with ready messages causing issues with Model Studio pipelines.
+
+            Click on the  URL on how to remediate the issue."
```
gsmith-sas, line 15: Remove quotes around description.  Remove line "Click on the URL on how to remediate the issue", I suspect it pointed to an internal runbook.  I think I would remove the reference to Model Studio as well.  That is not present in every SAS Viya deployment and I suspect/assume this alert is useful outside of the Model Studio use-case.

monitoring/alerting/rules/other_alerts.yaml
```diff
@@ -0,0 +1,123 @@
...
+    folder: Other Alerts
+    orgId: 1
+    rules:
+      - title: Crunchy Backrest Repo
+        annotations:
+          description:
+            "Checks to see /pgbackrest/repo1 filesystem is more than 50% full.
+
+            Go to the  URL to follow the troubleshooting steps."
+          summary:
```
gsmith-sas, line 14: Remove quotes in DESCRIPTION.  Remove reference to URL.  Why is this in Other Alerts instead of Database Alerts with the other Crunchy-related Alerts?

monitoring/alerting/rules/viya_platform_alerts.yaml
```diff
@@ -0,0 +1,324 @@
...
+        isPaused: false
+        labels: {}
+        noDataState: NoData
+        uid: e45e6d74-e396-40ce-a061-2a294295e61b
+      - title: RabbitMQ Ready Queue Backlog
+        annotations:
+          description:
+            Checks for accumulation of Rabbitmq ready messages > 10,000. It
+            could impact Model Studio pipelines. Follow the steps in the  url to help
+            troubleshoot. The  covers potential orphan queues and/or bottlenecking of
```
gsmith-sas, line 104: Again, remove mentions of URLs (internal runbooks) and Model Studio.

gsmith-sas (changes requested):
a few nits.

PR #850, opened 2026-04-15
change how datasources are configured for tempo to match standard

ceelias:


monitoring/bin/deploy_monitoring_openshift.sh
```diff
@@ -391,8 +392,10 @@ if [ "$TRACING_ENABLE" == "true" ]; then
     helm upgrade --install v4m-tempo \
         -n "$MON_NS" \
         -f "$imageKeysFile" \
-        -f monitoring/openshift/tempo-values.yaml \
+        -f monitoring/values-tempo.yaml \
         -f "$TEMPO_USER_YAML" \
+        --set "tempo.metricsGenerator.enabled=false" \
```
ceelias, line 396: set to false because the original open shift config had it set to false and Im unsure if its necessary. I'll likely remove when I get to the open shift story

monitoring/openshift/tempo-values.yaml
```diff
@@ -14,7 +14,7 @@ tempo:
   metricsGenerator:
     # -- If true, enables Tempo's metrics generator (https://grafana.com/docs/tempo/next/metrics-generator/)
     enabled: false
-    remoteWriteUrl: "http://prometheus.monitoring:9090/api/v1/write"
+    remoteWriteUrl: "http://v4m-prometheus.monitoring:9090/api/v1/write"  # overridden at deploy time with MON_NS
```
ceelias, line 17: Nolonger points to this file, but will leave for now. Will probs remove when I can test everything properly on open shift

monitoring/values-tempo.yaml
```diff
@@ -20,7 +20,7 @@ tempo:
   metricsGenerator:
     # -- If true, enables Tempo's metrics generator (https://grafana.com/docs/tempo/next/metrics-generator/)
     enabled: true
-    remoteWriteUrl: "http://v4m-prometheus.monitoring:9090/api/v1/write"
+    remoteWriteUrl: "http://v4m-prometheus.monitoring:9090/api/v1/write"  # overridden at deploy time with MON_NS
```
gsmith-sas: Probably better to remove this line entirely.  It will default to whatever has been identified as the default value in the Helm chart.

gsmith-sas (changes requested):

monitoring/openshift/tempo-values.yaml
gsmith-sas, line 17: We should remove this but I also know that we have additional tracing on OpenShift worked planned. So, I guess it can wait until then.

gsmith-sas (approved):
See note re:OpenShift, but no need to hold things up further so approving this PR.

monitoring/openshift/tempo-values.yaml
ceelias: yes, as mentioned, this entire file should theoretically be obsolete. I'll remove it once I get to the open shift story

PR #859, opened 2026-05-13
Unified grafana monitoring stack

SimonasBanys:
Deploying unified grafana monitoring stack for metrics, logs, and traces, utilising open telemetry correlation for observability and enabling issue investigations


gsmith-sas:
@SimonasBanys  Thank you for submitting this PR to add support for the Grafana LGTM stack to the project. We will need some time to review the PR, get a better understanding of these changes and how they fit into the project's overall roadmap.  During that process we may reach out to you for more information or to suggest changes.  Thank you once again for taking the time to craft a fix and submitting it.

ceelias:
Hey, @SimonasBanys thanks so much for putting this together. This is a significant amount of work and it's genuinely appreciated. The move to a unified LGTM stack with Alloy-based collection, native Loki integration, and full trace correlation is a direction we have discussed as a team in the past but have not decided on.

Given the scope of the change, replacing the core monitoring stack, introducing several new components, and touching nearly every deployment artifact, we need to take some time to evaluate whether this is the direction we want to take the project before committing to it. This is a substantial architectural shift and we want to make sure we're aligned as a team before merging something of this size.

With that in mind, we'll be closing this PR. If we decide to move forward in this direction, we'll reference this PR and work. We just don't want to leave a large open PR in an uncertain state while we work through the decision.
That said, we did spin this up and put it through some review and overall it looks great and largely works as described. In the spirit of transparency, incase others find this PR/branch and try to use it, we wanted to call out a few issues we encountered for posterity:
1. *macOS compatibility:* `sed -i` without an empty-string argument breaks on BSD/macOS and needs `sed -i ''`
1. *TLS cert generation removed:* the previous stack auto-generated self-signed certs via `tls-include.sh` but that path has been dropped, leaving users without cert-manager in a broken ingress state when `BASE_DOMAIN` is set
1. *Air-gap support broken:* the script now fetches the Alloy Operator CRD directly from GitHub at deploy time, which breaks offline/air-gapped deployments the previous version explicitly supported
1. *`user.env` contains test values:* `MON_NS`, `HELM_RELEASE_PREFIX`, and `BASE_DOMAIN` are set to environment-specific values and shouldn't be committed
1. *Uninstall ordering:* `remove_monitoring_cluster.sh` removes Prometheus and Loki before k8s-monitoring (Alloy) and Alloy should be torn down first to avoid backpressure errors during teardown
1. *Alloy receiver ConfigMap bypasses Helm:* the hand-crafted ConfigMap carries Helm annotations but is applied via `kubectl apply`, meaning it will be silently overwritten on the next `helm upgrade`

Thanks again for the contribution. This is solid work and we always welcome input and changes from the community!

PR #879, opened 2026-07-09
Semver cleanup

gsmith-sas:


bin/common.sh
```diff
@@ -37,22 +37,231 @@ function errexit_msg {
...
+    ## portions of the semantic version string passed to it
+    ##
+    ## If the passed string is NOT a valid semantic version string,
+    ## this function returns nothing and sets a non-zero exit code
+
+    local string fragment
+    string=$1
+    fragment=${2:-ECHO}
+
+    if [[ $string =~ v?([0-9]+)\.([0-9]+)\.([0-9]+)(-(([0-9]+|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.([0-9]+|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$ ]]; then
```
gsmith-sas: Technically, this regex is not completely correct: it permits leading zeros which are not permitted in the SemVer spec. ☹️  However, this regex is 155 chars long and the fully correct one extends to 191 characters.  Willing to "correct" this if needed.

bin/kube-include.sh
```diff
@@ -11,54 +11,44 @@ if [ ! "$(which kubectl)" ]; then
...
+# 2024.03 LTS 1.26 1.28   (EOL: 1.26)
+# 2024.09 LTS 1.28 1.30   (EOL: 1.28)
+# 2025.03 LTS 1.29 1.31
+# 2025.09 LTS 1.31 1.33
+# 2026.01     1.31 1.33
+# 2026.02     1.32 1.34
+# 2026.03     1.32 1.34
+# 2026.04     1.32 1.34
+
+KUBE_MIN_VER=${KUBE_MIN_VER:-"1.28.0"} #TO DO: Keep this changeable via env var?
```
gsmith-sas: I realized this morning that we should be setting this (and the other "minimum version required" settings)  in the `component_versions.env` file (i.e. where we keep track of all of our other versioning details)!  Will take care of that on Monday.

gsmith-sas: Done!

bin/common.sh
gsmith-sas: Revised to support full SemVer spec inc. prohibition on leading zeros.

ceelias (approved):

PR #891, opened 2026-08-21
KSM Rbac Proxy and Network Policy

ceelias:
## Summary by CodeRabbit

* **Security**
  * Added an optional RBAC proxy to protect Kube State Metrics and authenticate Prometheus scraping.
  * Restricted Kube State Metrics access to Prometheus pods by default.
  * Added support for custom NetworkPolicy rules and HTTPS scraping through the proxy.

* **Documentation**
  * Documented RBAC proxy settings, NetworkPolicy customization, and contributor workflow updates.

* **Chores**
  * Added proxy image details to deployment and artifact inventories.
  * Updated the chart version.


gsmith-sas:
- The `components_versions.env` file needs to be updated to include a reference to the KSM RBAC Proxy image.  This will ensure it is reflected in the `ARTIFACTS_INVENTORY.md` file and folks deploying into an air-gap environment can add it to their registry.  

- How the container image is specified might also need to be changed to ensure it is handled consistently with the other container images and our air-gap deployment processing.

ceelias:
Closing as these changes have been grouped with the KSM changes into another branch/pr: #896

PR #895, opened 2026-08-31
Node exporter data leakage

ceelias:


.github/workflows/build-artifact/ARTIFACT_INVENTORY.template
```diff
@@ -26,6 +26,7 @@ registry/repository/image_name:version
 | Metrics | [Admission Webhook](https://github.com/kubernetes/ingress-nginx) | __ADMWEBHOOK_FULL_IMAGE__ |
 | Metrics | [Kube State Metrics](https://github.com/kubernetes/kube-state-metrics) | __KSM_FULL_IMAGE__ |
 | Metrics | [Node Exporter](https://github.com/prometheus/node_exporter) | __NODEXPORT_FULL_IMAGE__ |
+| Metrics | [Kube RBAC Proxy](https://github.com/brancz/kube-rbac-proxy) | __RBAC_PROXY_FULL_IMAGE__ |
```
gsmith-sas, line 29: I think the URL here should point to the Kube RBAC Proxy project, i.e. https://github.com/kube-rbac-proxy/kube-rbac-proxy, rather than the image.

component_versions.env
```diff
@@ -53,6 +53,7 @@ ALERTMANAGER_FULL_IMAGE="quay.io/prometheus/alertmanager:v0.32.1"
 ADMWEBHOOK_FULL_IMAGE="ghcr.io/jkroepke/kube-webhook-certgen:1.8.2"
 KSM_FULL_IMAGE="registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.19.1"
 NODEXPORT_FULL_IMAGE="quay.io/prometheus/node-exporter:v1.11.1-distroless"
+RBAC_PROXY_FULL_IMAGE="quay.io/brancz/kube-rbac-proxy:v0.22.0"
```
gsmith-sas, line 56: Looks like the latest release is v0.22.1.  Any specific reason/concern that led to choosing the earlier v0.22.0 release?

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -635,6 +641,22 @@ fi
...
 
+# node-exporter NetworkPolicy: restrict ingress to Prometheus pods only.
+# Applied unconditionally (independent of RBAC_PROXY_ENABLE) since
+# NetworkPolicies are additive -- this can only ever add an allowed ingress
+# source, never weaken or replace a site's own existing policies. A site that
+# wants custom ingress rules, or hits a naming collision with a NetworkPolicy
+# they already manage, can drop their own replacement at
+# $USER_DIR/monitoring/networkPolicy-node-exporter.yaml.
+# NOTE: Node Exporter runs with hostNetwork: true, so this NetworkPolicy is
+# not enforced on all CNIs (confirmed on Calico); actual access control is
```
gsmith-sas, line 652: This explanation is very helpful.  However, this last note is a bit unclear.  Does "_confirmed on Calico_" mean that you confirmed that the Network Policy is _not_ enforced on Calico?  Or, that you were able to confirm the Network Policy _was enforced_ on Calico?

monitoring/tls/values-prom-operator-tls.yaml
```diff
@@ -38,20 +38,6 @@ alertmanager:
     tlsConfig:
       insecureSkipVerify: true
 
-prometheus-node-exporter:
-  extraArgs:
```
gsmith-sas, line 42: With this removed, what happens if someone disables the RBAC Proxy AND still wants TLS enabled?

samples/generic-base/monitoring/networkPolicy-node-exporter.yaml
```diff
@@ -0,0 +1,31 @@
...
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - podSelector:
+        matchLabels:
+          app.kubernetes.io/name: prometheus
+    ports:
+    - protocol: TCP
+      port: 9110
```
gsmith-sas, line 31: The contents of all of the other files in the `samples/generic-base` directory are commented out.  They are intended to show users environment variables they might want to set and, generally, their default values.  Or they provide similar information for the user value yaml files consumed by the Helm charts.  If this file is intended to serve a similar role, its contents should be commented out as well.  But I don't think we envision users selectively uncommenting portions of this file...or even using this file as-is.

I think this case (i.e. NetworkPolicy) is more similar to the cases of the Grafana Dashboards and Grafana Alert content in this sub-directory.  In those cases, we include a `README.md`  that explains how the user could add their own yaml/json files to define those resources and we will pick them up during the deployment process.  And we do not include any samples of those types of files.  I think that's the appropriate model to follow in this case.  In those README.md files, we point to corresponding information in our Help Center documentation.  We don't have anything in our Help Center doc about NetworkPolicy (yet?), so I think it would be appropriate to point to the Kubernetes documentation for the NetworkPolicy resource.  Our existing NetworkPolicy is already defined elsewhere in our repo.  If the user has a need to define a new/different NetworkPolicy, it would be better for them to read the K8s document rather than incrementally alter our existing definition.  

In addition, our documentation suggests users copy the generic-base sample to create their initial USER_DIR directory.  Keeping this file in its current form means that it would be used rather than our official NetworkPolicy when they point to their USER_DIR.  That's fine now when the files are the same...but introduces problems if/when we make a change to the official NetworkPolicy in the future and the instance in their existing USER_DIR becomes obsolete.

This probably means we need a `samples/generic-base/monitoring/networkPolicy` sub-directory containing a README.md file.  AND, logic in `deploy_monitoring_cluster.sh` is updated to pull a user's NetworkPolicy yaml file from `$USER_DIR/monitoring/networkPolicy/networkPolicy-node-exporter.yaml` rather than `$USER_DIR/monitoring/networkPolicy-node-exporter.yaml`.   Hmmm, _might_ want to name the actual yaml file `node-exporter.yaml` since the sub-directory name indicates it is a networkPolicy.

samples/generic-base/README.md
```diff
@@ -30,3 +30,12 @@ log-monitoring components. See [Deploy](https://documentation.sas.com/?softwareI
...
 [Provision Alert Rules, Contact Points, and Notification Policies](https://documentation.sas.com/?softwareId=obsrv&softwareVersion=prod&docsetId=obsrvdply&docsetTarget=n0auhd4hutsf7xn169hfvriysz4e.htm#p1cb39v8tr2379n1c4posre4v8wx) for details.
+
+## Node Exporter NetworkPolicy
+
+The `monitoring/networkPolicy-node-exporter.yaml` file in this sample is an
+optional override for the default NetworkPolicy that restricts access to the
+Node Exporter service to Prometheus pods only. This default NetworkPolicy is
+always applied automatically; copy this file to your customization directory
+only if you need custom ingress rules, or if you already manage a
+NetworkPolicy of your own with the same name in the monitoring namespace.
```
gsmith-sas, line 41: See comments above.  Replace this with something shorter...probably the same or similar content to what is put into the (new) `samples/generic-base/monitoring/networkPolicy/README.md` file.

CHANGELOG.md
```diff
@@ -8,6 +8,17 @@ is configured automatically
...
+kernel info) over an unauthenticated, unencrypted HTTP endpoint reachable by any pod in the cluster.
+Node Exporter is now fronted by a `kube-rbac-proxy` sidecar enforcing Kubernetes RBAC and bound to
+localhost only; Prometheus scrapes it using its existing ServiceAccount token, with no certificate
+management required. This is controlled by the new `RBAC_PROXY_ENABLE` environment variable (default
+`true`), which is independent of `TLS_ENABLE`. A NetworkPolicy restricting access to Prometheus pods
+only is also always applied, regardless of `RBAC_PROXY_ENABLE`, since NetworkPolicies are additive and
+cannot weaken or replace a site's existing policies; sites needing custom rules or hitting a naming
+collision can supply their own at `$USER_DIR/monitoring/networkPolicy-node-exporter.yaml`. Because Node
+Exporter runs with `hostNetwork: true`, this NetworkPolicy is not enforced on all CNIs (confirmed on
+Calico); actual access control is provided by the `kube-rbac-proxy` RBAC check.
```
gsmith-sas, line 21: Split this in two.  Since NetworkPolicy is implemented separately and is not disabled by the `RBAC_PROXY_ENABLE` env var, there's no reason to combine the description.  Breaking it out will make it much more clear that we have introduced this 2nd, additional, security feature.

I wonder if the last sentence should be reworded to say "_Because Node
Exporter runs with `hostNetwork: true`, this NetworkPolicy may not be enforced on all Container Network Interfaces (CNI)._"  In addition, if we know it will not be enforced on Calico, add another sentence "_For example, we have confirmed this NetworkPolicy will **not** be enforced with the Calico CNI._"

With regard to the last part of the sentence (i.e. "_actual access control is provided by the `kube-rbac-proxy` RBAC check._"), I think the point we're trying to make is that even if the NetworkPolicy is NOT enforced, the proxy RBAC check will still provide protection.  If so, maybe add it as a final sentence in the NetworkPolicy entry. something like: "_Even if the NetworkPolicy is not enforced by your CNI, the RBAC Proxy, if enabled, will still limit access to the Node Exporter pods to the Prometheus ServiceAccount._"

gsmith-sas (changes requested):
This looks good.  However, I do think some changes are required, see comments for specifics.

ceelias:
Closing as these changes have been grouped with the KSM changes into another branch/pr: #896

PR #896, opened 2026-09-03
Data Leakage Fix for KSM and Node Exporter using Kube RBAC Proxy and Network Policy

ceelias:
## Summary by CodeRabbit

- **Security**
  - Added Kubernetes RBAC proxy sidecars for Prometheus, Kube State Metrics, and Node Exporter.
  - Metrics scraping now uses authenticated HTTPS connections through the proxies.
  - Retired Node Exporter’s native TLS configuration.

- **Configuration**
  - Added an environment setting to enable or disable RBAC proxies, enabled by default.
  - Added deployment settings and artifact inventory support for the proxy image.
  - Grafana Prometheus queries now use authenticated bearer tokens when proxies are enabled.

- **Behavior Changes**
  - Removed NetworkPolicy-based access restrictions for Kube State Metrics and Node Exporter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,17 @@
 # SAS Viya Monitoring for Kubernetes
+## Unreleased
+* **Metrics**
+  * [SECURITY] Kube State Metrics (KSM) and Node Exporter previously exposed cluster-wide resource state
+and infrastructure metrics -- including pod labels used to identify per-user SAS job activity -- over
+unauthenticated, unencrypted HTTP endpoints reachable by any pod in the cluster. Both are now fronted by
+a `kube-rbac-proxy` sidecar enforcing Kubernetes RBAC and bound to localhost only; Prometheus scrapes
+them using its existing ServiceAccount token, with no certificate management required. This is
+controlled by the new `RBAC_PROXY_ENABLE` environment variable (default `true`), which is independent
+of `TLS_ENABLE`. `kube-rbac-proxy` is also the only source of transport encryption for both targets;
+setting `RBAC_PROXY_ENABLE=false` serves them over plain, unauthenticated HTTP regardless of
+`TLS_ENABLE`. Node Exporter previously supported its own native TLS independent of RBAC/auth; that
+capability has been retired in favor of `kube-rbac-proxy`.
+
 ## Version 1.2.54 (04SEP2026)
 * **Overall**
   * [CHORE] Refactor Helm Chart version checking/reporting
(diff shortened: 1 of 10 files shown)


monitoring/tls/values-prom-operator-tls.yaml
```diff
@@ -38,20 +38,6 @@ alertmanager:
     tlsConfig:
       insecureSkipVerify: true
 
-prometheus-node-exporter:
```
ceelias, line 41: Responding to other PR where Greg asked what this changes. Looks like node-exporter-web.yaml was always commented out so it never actually enabled TLS for node exporter anyway.

monitoring/values-prom-operator.yaml
```diff
@@ -212,6 +234,24 @@ prometheus-node-exporter:
...
+    ingress:
+      - from:
+          - podSelector:
+              matchLabels:
+                app.kubernetes.io/name: prometheus
+        ports:
+          - protocol: TCP
+            port: 9110
+    egress:
+      - {}
```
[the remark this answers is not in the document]
ceelias: Hmm, this seems like an interesting issue. The helm chart basically denies all egress when networkpolicy is enabled and it can't be overwritten which would cause failures. I didn't catch this because it depends on how the CNI the network policy against hostnetworking pods. So if deployed in an env with a CNI that does enforce it, this would cause in issue. The previous solution wouldn't have this issue as we could write the egress, but in this case the helm chart doesn't support that

monitoring/bin/deploy_monitoring_cluster.sh
```diff
@@ -524,6 +535,52 @@ else
         --dry-run=client -o yaml | kubectl apply -f -
 fi
 
+ksmProbePatchPid=""
+if [ "$RBAC_PROXY_ENABLE" == "true" ]; then
+    # Work around an upstream kube-state-metrics chart bug: when
```
gsmith-sas, line 540: Let's add a link to the bug/issue.  At some point, it may be fixed and we want to be able to check for that periodically.

CHANGELOG.md
```diff
@@ -22,6 +22,17 @@ is configured automatically
...
+and infrastructure metrics -- including pod labels used to identify per-user SAS job activity -- over
+unauthenticated, unencrypted HTTP endpoints reachable by any pod in the cluster. Both are now fronted by
+a `kube-rbac-proxy` sidecar enforcing Kubernetes RBAC and bound to localhost only; Prometheus scrapes
+them using its existing ServiceAccount token, with no certificate management required. This is
+controlled by the new `RBAC_PROXY_ENABLE` environment variable (default `true`), which is independent
+of `TLS_ENABLE`. `kube-rbac-proxy` is also the only source of transport encryption for both targets;
+setting `RBAC_PROXY_ENABLE=false` serves them over plain, unauthenticated HTTP regardless of
+`TLS_ENABLE`. Node Exporter previously supported its own native TLS independent of RBAC/auth; that
+capability has been retired in favor of `kube-rbac-proxy`. Not applicable to OpenShift, which does not
+deploy KSM or Node Exporter as part of this project.
```
gsmith-sas: No need to mention OpenShift here.  While true, the entry is long enough w/o it and we don't need to mention it every time we talk about metrics.

gsmith-sas (approved):
A couple of minor tweaks to comments and the CHANGELOG.md.  Approving with the expectation that they will be addressed prior to merging the PR.

monitoring/bin/deploy_monitoring_cluster.sh
ceelias: Its linked at the bottom of the comment

ceelias:
Just for transparency, I force pushed to the branch in order to rebase off of main and cleanup all the commits. Now it should only have a few commits

monitoring/values-prom-operator-rbac-proxy.yaml
```diff
@@ -0,0 +1,183 @@
...
+  prometheus:
+    monitor:
+      # kube-rbac-proxy terminates TLS with a self-signed cert (no cert
+      # management required) and authenticates the scrape using Prometheus's
+      # own ServiceAccount token.
+      http:
+        scheme: https
+        tlsConfig:
+          insecureSkipVerify: true
+        bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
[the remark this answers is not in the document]
vishaanS: Not sure if we can do this given our current architecture?

gsmith-sas (approved):
Re-reviewed after the addition of the Prometheus enhancements.  LGTM

PR #901, opened 2026-09-09
[FIX] Only delete HTTPProxy resource if CRD exists

gsmith-sas:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging cleanup to avoid attempting to remove the root HTTPProxy when the required Contour resource definition is unavailable.
  * Root HTTPProxy creation now occurs only when ingress autogeneration, root-proxy creation, and Contour ingress are enabled.
  * Applied safer handling across supported ingress configuration scenarios.

* **Documentation**
  * Updated the changelog with details of the logging cleanup fix and root HTTPProxy behavior.

### DIFF
--- a/logging/bin/deploy_logging.sh
+++ b/logging/bin/deploy_logging.sh
@@ -61,7 +61,6 @@ if [ "$AUTOGENERATE_INGRESS" == "true" ]; then
     else
         if kubectl get crd httpproxies.projectcontour.io -o name > /dev/null 2>&1; then
 
-
             log_debug "Deleting [httpproxy/v4m-logging-root-proxy] if it exists"
             kubectl -n "$LOG_NS" delete httpproxy v4m-logging-root-proxy --ignore-not-found
         fi
--- a/logging/bin/deploy_logging.sh
+++ b/logging/bin/deploy_logging.sh
@@ -63,8 +63,10 @@ if [ "$AUTOGENERATE_INGRESS" == "true" ]; then
         kubectl -n "$LOG_NS" delete httpproxy v4m-logging-root-proxy --ignore-not-found
     fi
 else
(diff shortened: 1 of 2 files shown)


logging/bin/deploy_logging.sh
```diff
@@ -59,12 +59,17 @@ if [ "$AUTOGENERATE_INGRESS" == "true" ]; then
     if [ "$INGRESS_CREATE_ROOT_PROXY" == "true" ] && [ "$INGRESS_TYPE" == "contour" ]; then
         create_root_httpproxy logging "$LOG_NS"
     else
-        log_debug "Deleting [httpproxy/v4m-logging-root-proxy] if it exists"
-        kubectl -n "$LOG_NS" delete httpproxy v4m-logging-root-proxy --ignore-not-found
+        if kubectl get crd httpproxies.projectcontour.io -o name > /dev/null 2>&1; then
```
ceelias: Not a big deal but this if else block looks a bit confusing and if I'm following the logic correctly I think it could be simplified to something like:


```bash
if [[ "$AUTOGENERATE_INGRESS" == "true" &&
      "$INGRESS_CREATE_ROOT_PROXY" == "true" &&
      "$INGRESS_TYPE" == "contour" ]]; then

    create_root_httpproxy logging "$LOG_NS"

elif kubectl get crd httpproxies.projectcontour.io -o name > /dev/null 2>&1; then
    log_debug "Deleting [httpproxy/v4m-logging-root-proxy] if it exists"
    kubectl -n "$LOG_NS" delete httpproxy v4m-logging-root-proxy --ignore-not-found
fi
```

gsmith-sas: Changed as you suggested.

ceelias (approved):

